PR #1, opened 2016-03-19
MVC_SPEC-59 Added support for new ValidationError API

chkal:
This PR contains the required changes for Ozark to support the new `ValidationError` interface and the modified `BindingResult`.

Please note the current version of the implementation doesn't yet set the parameter name correctly. This has to be done in a later step. But IMO that's ok for now as currently Ozark also doesn't know about the parameter name and therefore even doesn't implement `BindingResult.getViolation(String)` at all.

### DIFF
--- a/ozark/src/main/java/org/glassfish/ozark/binding/BindingResultImpl.java
+++ b/ozark/src/main/java/org/glassfish/ozark/binding/BindingResultImpl.java
@@ -42,11 +42,13 @@ package org.glassfish.ozark.binding;
 import javax.enterprise.context.RequestScoped;
 import javax.mvc.binding.BindingError;
 import javax.mvc.binding.BindingResult;
-import javax.validation.ConstraintViolation;
+import javax.mvc.binding.ValidationError;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * Implementation for {@link javax.mvc.binding.BindingResult} interface.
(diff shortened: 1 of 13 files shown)


ozark/src/main/java/org/glassfish/ozark/binding/BindingResultImpl.java
```diff
@@ -89,22 +91,26 @@ public BindingError getBindingError(String param) {
...
 
     @Override
-    public Set<ConstraintViolation<?>> getAllViolations() {
-        return violations;
+    public Set<ValidationError> getAllValidationErrors() {
+        return Collections.unmodifiableSet(validationErrors);
     }
 
     @Override
-    public Set<ConstraintViolation<?>> getViolations(String propertyPath) {
```
ivargrimstad: Change parameter name to param, paramName?

ozark/src/main/java/org/glassfish/ozark/binding/BindingResultImpl.java
```diff
@@ -89,22 +91,26 @@ public BindingError getBindingError(String param) {
...
-        throw new UnsupportedOperationException("Not supported");
+    public Set<ValidationError> getValidationErrors(String param) {
```
ivargrimstad: See comment above

ozark/src/main/java/org/glassfish/ozark/binding/BindingResultImpl.java
```diff
@@ -89,22 +91,26 @@ public BindingError getBindingError(String param) {
...
+        return validationErrors.stream()
+                .filter(ve -> Objects.equals(ve.getParamName(), param))
+                .collect(Collectors.toSet());
     }
 
     @Override
-    public ConstraintViolation<?> getViolation(String propertyPath) {
-        throw new UnsupportedOperationException("Not supported");
```
ivargrimstad: See comment above

ozark/src/main/java/org/glassfish/ozark/binding/BindingResultImpl.java
chkal: Good catch. I'll update these. Seems like they are out of sync with the interface for quite some time. It was called `propertyPath` in the old days. :)

chkal: I checked this again. Looks like it was called "propertyPath" only in the old version of the file. My patch seems to be fine and is using "param" instead.

PR #2, opened 2016-03-29
Implementation of MVC_SPEC-27 "Handling of Locale and Language"

chkal:
Full implementation of this proposal:

https://github.com/mvc-spec/mvc-spec/pull/2

### DIFF
--- a/ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
+++ b/ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
@@ -43,7 +43,6 @@ import javax.annotation.Priority;
 import javax.enterprise.context.ApplicationScoped;
 import javax.mvc.locale.LocaleResolver;
 import javax.mvc.locale.LocaleResolverContext;
-import javax.ws.rs.core.HttpHeaders;
 import java.util.Locale;
 
 /**
@@ -58,15 +57,13 @@ public class DefaultLocaleResolver implements LocaleResolver {
     @Override
     public Locale resolveLocale(LocaleResolverContext context) {
 
-        /*
-         * We check if there is a "Accept-Language" header sent. If so, we can use
-         * HttpServletRequest#getLocale() to get the locale. We must check for the header
-         * because HttpServletRequest#getLocale() returns the system default locale if
-         * the client didn't send this header.
-         */
-        String header = context.getRequest().getHeader(HttpHeaders.ACCEPT_LANGUAGE);
-        if (header != null && header.trim().length() > 0) {
-            return context.getRequest().getLocale();
+        // Prefer language from "Accept-Language" header
+        Locale clientLocale = context.getAcceptableLanguages().stream()
+                .filter(l -> !"*".equals(l.getLanguage()))
+                .findFirst()
+                .orElse(null);
+        if (clientLocale != null) {
+            return clientLocale;
         }
 
         // Lookup application default locale
--- a/ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
+++ b/ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
@@ -43,6 +43,7 @@ import javax.annotation.Priority;
 import javax.enterprise.context.ApplicationScoped;
 import javax.mvc.locale.LocaleResolver;
 import javax.mvc.locale.LocaleResolverContext;
+import javax.ws.rs.core.HttpHeaders;
 import java.util.Locale;
 
 /**
@@ -63,7 +64,7 @@ public class DefaultLocaleResolver implements LocaleResolver {
          * because HttpServletRequest#getLocale() returns the system default locale if
          * the client didn't send this header.
          */
-        String header = context.getRequest().getHeader("Accept-Language");
+        String header = context.getRequest().getHeader(HttpHeaders.ACCEPT_LANGUAGE);
         if (header != null && header.trim().length() > 0) {
             return context.getRequest().getLocale();
         }
--- a/ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
+++ b/ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
@@ -37,77 +37,35 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-package org.glassfish.ozark;
+package org.glassfish.ozark.locale;
 
-import javax.enterprise.context.RequestScoped;
-import javax.inject.Inject;
-import javax.inject.Named;
-import javax.mvc.MvcContext;
-import javax.mvc.security.Csrf;
-import javax.mvc.security.Encoders;
-import javax.ws.rs.core.Configuration;
-
+import javax.annotation.Priority;
+import javax.enterprise.context.ApplicationScoped;
+import javax.mvc.locale.LocaleResolver;
+import javax.mvc.locale.LocaleResolverContext;
 import java.util.Locale;
 
 /**
- * Implementation of {@link javax.mvc.MvcContext}.
+ * Default implementation of {@link LocaleResolver} as defined in the spec.
  *
- * @author Santiago Pericas-Geertsen
+ * @author Christian Kaltepoth
  */
-@Named("mvc")
-@RequestScoped
-public class MvcContextImpl implements MvcContext {
-
-    @Inject
-    private Csrf csrf;
-
-    @Inject
-    private Encoders encoders;
-
-    @Inject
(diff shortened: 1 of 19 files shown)


ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+@ApplicationScoped
+@Priority(0)
+public class DefaultLocaleResolver implements LocaleResolver {
+
+    @Override
+    public Locale resolveLocale(LocaleResolverContext context) {
+
+        // Get locale from the "Accept-Language" header
+        Locale locale = context.getRequest().getLocale();
```
ivargrimstad: See my comment on the spec PR

ozark/src/main/java/org/glassfish/ozark/locale/LocaleResolverChain.java
```diff
@@ -0,0 +1,126 @@
...
+    @Context
+    private Configuration configuration;
+
+    @PostConstruct
+    public void verify() {
+        Objects.requireNonNull(configuration, "The Configuration instance was not injected! " +
+                "Please make sure you are using a recent version of Jersey.");
+    }
+
+    public Locale resolve() {
```
ivargrimstad: If you change this method signature to be

`public Local resolve(ContainerRequestContext requestContext)`

Then you will have access to all information and won't need to inject the `HttpServletContext`.

Set the `ContainerRequestContext` on the `LocalResolverContext` (API needs to be updated for this)

chkal: I agree that `LocaleResolverChain.resolve()` should get the `ContainerRequestContext` as a parameter as this reduces the need to inject things into this class.

ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public Locale resolveLocale(LocaleResolverContext context) {
+
+        /*
+         * We check if there is a "Accept-Language" header sent. If so, we can use
+         * HttpServletRequest#getLocale() to get the locale. We must check for the header
+         * because HttpServletRequest#getLocale() returns the system default locale if
+         * the client didn't send this header.
+         */
+        String header = context.getRequest().getHeader("Accept-Language");
```
lefloh: Maybe just a preference but could we use a constant like `javax.ws.rs.core.HttpHeaders.ACCEPT_LANGUAGE`?

chkal: I agree. We should use `HttpHeaders.ACCEPT_LANGUAGE` here.

ozark/src/main/java/org/glassfish/ozark/locale/LocaleResolverChain.java
chkal: Just fixed that one...

ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
chkal: I just fixed that one.

ozark/src/main/java/org/glassfish/ozark/locale/LocaleResolverChain.java
ivargrimstad: I still can't entirely let it go why we can't use `ContainerRequestContext` all the way. If we change the LocalResolverContext interface of the API to this:
```
public interface LocaleResolverContext {

    ContainerRequestContext getRequest();
    Configuration getConfiguration();
}
```
Then the `resolveLocal` method of `DefaultLocaleResolver` will be something like this:
```
      String header = context.getRequest().getHeaderString(HttpHeaders.ACCEPT_LANGUAGE);
        if (header != null && header.trim().length() > 0) {
            return context.getRequest().getLanguage();
        }
```
Sorry for being difficult :)

lefloh: I think the main reason is the documentation: 

> A mutable class that provides request-specific information for the filter...

The `ContainerRequestContext` is tied to a `ContainerRequestFilter`. Also I'm not sure if we should provide `ContainerRequestContext#abortWith` while resolving the locale?

ivargrimstad: Good point! Thanks!

Then we just have to provide whatever information in the context that isn't directly injectable or possible to obtain for the implementers. 

For example, for the current implementation, only the configuration is needed since `HttpServletContext` may be injected as discussed with Christian in the MVC PR.

chkal: I also agree with @lefloh. I think we should only expose things that are useful to the user though `LocaleResolverContext`. On the Ozark side we can basically let `LocaleResolverContextImpl` wrap a `ContainerRequestContext` and delegate relevant calls to that instance.

ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.ws.rs.core.HttpHeaders;
+import java.util.Locale;
+
+/**
+ * Default implementation of {@link LocaleResolver} as defined in the spec.
+ *
+ * @author Christian Kaltepoth
+ */
+@ApplicationScoped
+@Priority(0)
```
Turini: What do you think about using `Priority.LIBRARY_BEFORE` instead?
and users may override this using `Priority.APPLICATION`

ozark/src/main/java/org/glassfish/ozark/locale/LocaleResolverChain.java
```diff
@@ -0,0 +1,126 @@
...
+        for (LocaleResolver candidate : candidates) {
+            Locale locale = candidate.resolveLocale(context);
+            if (locale != null) {
+                return locale;
+            }
+        }
+
+        throw new IllegalStateException("Could not resolve locale with any of the " + candidates.size()
+                + " resolver implementations");
+
```
Turini: it may be a bit more human readable extracting a few methods and exploring a bit more the streams api:
```
StreamSupport.stream(resolvers.spliterator(), false)
    .sorted(this::byPriority)
    .map(candidate -> candidate.resolveLocale(context))
    .filter(candidate -> candidate != null)
    .findFirst()
    .orElseThrow(() -> new IllegalStateException("Could not resolve ..."));
```
what do you think?

note: see my comments about optional on mvc-spec https://github.com/mvc-spec/mvc-spec/pull/2/files#r58268281

ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
chkal: This may be a dump question, but which `Priority.LIBRARY_BEFORE` are you referring to?

We already have a `Priorities` class in MVC, but the javadocs explicitly refer to `ViewEngine`, that's why I didn't use constants from this.

ozark/src/main/java/org/glassfish/ozark/locale/LocaleResolverChain.java
chkal: I agree that this method could be implemented in a nicer way. So +1 for changing that. But that's implementation detail. The pull request for Ozark is mainly about showing that my API proposal can be implemented in the RI and for providing a first working version. We can clean up the implementation after everything is merged. Is that ok?

ozark/src/main/java/org/glassfish/ozark/locale/DefaultLocaleResolver.java
Turini: I'm talking about existing CDI's priority range convention
https://javaee-spec.java.net/nonav/javadocs/javax/interceptor/Interceptor.Priority.html
It's widest used to define the priority of CDI beans (interceptors, alternatives, ...)
not sure if it fits well here, actually. What do you think?

ozark/src/main/java/org/glassfish/ozark/locale/LocaleResolverChain.java
Turini: > We can clean up the implementation after everything is merged. Is that ok?

Yes, definitively! 
I can help with this, if you want. Just let me know or fill an issue assigning me :)

chkal: Awesome. Thanks a lot. :)

PR #4, opened 2016-04-03
Package MVC 1.0 RI as bundle

rahmanusta:


ozark/pom.xml
```diff
@@ -7,12 +7,53 @@
         <artifactId>parent</artifactId>
         <version>1.0.0-m03-SNAPSHOT</version>
     </parent>
-    <artifactId>ozark</artifactId>
-    <packaging>jar</packaging>
-    <name>Ozark ${project.version} MVC RI</name>
+    <artifactId>javax.mvc</artifactId>
```
chkal: You changed the artifact name from `ozark` to `javax.mvc`. I don't think that this is correct. ;)

rahmanusta: Hmm, I've seen that the all RIs uses this convention, for example look at Json-P RI https://java.net/projects/jsonp/sources/git/content/impl/pom.xml?rev=6bac828463f2b714effad5769d6f0a2bfc420af2

chkal: But this will change the Maven artifact coordinates. This will break all existing apps that declare Ozark as a dependency. Including the tests applications in the Ozark repository. So I think we should keep the artifact name as it is today.

Anyone has a different opinion on this?

rahmanusta: I think we can leave it "ozark" for this version but for the next version upgrade it should be javax.mvc for the convention. Anyone has different knowledge for the convention ?

lefloh: I don't think there's a convention at all. Mojarra uses `com.sun.faces:jsf-impl` and Jersey `org.glassfish.jersey.core:jersey-server`. The GAV [has already been changed](https://github.com/mvc-spec/ozark/commit/9c21a84f616e971dcd809d5a1c830bdf9a006dce) with [OZARK-51](https://java.net/jira/browse/OZARK-51) so I wouldn't change it again. Maybe you could start a discussion on the EG list?

rahmanusta: Hi @lefloh , Im convinced, yes there seems no convention. We can leave the artifact name same, but we can open a discussion on EG list if you want.

rahmanusta: Is it okay now ? @chkal @lefloh

chkal:
Hey @rahmanusta!

I just started to integrate this patch into the master branch. Unfortunately there were some conflicts and I also had to add some more CDI beans to the CDI extension which where added in the pull requests merged in the last days.

But now I'm getting integration test failures for tests related to the view engine extensions. I'll have to dig deeper into this. Hopefully today or at the weekend.

chkal:
I've some more details about this issue.

It looks like Ozark isn't picking up view engines from other CDI archives any more. They are simply not accessible via `@Inject @Any Instance<ViewEngine>`. That's why many tests are broken.

I did some more debugging on this. The Weld `BeanManagerImpl` class has a field `accessibleManagers`. Before applying your patch this list contains a reference to the bean manager of the corresponding Ozark extension JAR (`ozark-freemarker.war` in my case). After applying the patch, the list is smaller and doesn't contain the bean manager any more. There is also a field `managers` which has the same size before and after applying the patch.

Looks like some CDI vs OSGi issue here. Any idea?

If anyone wants to dig deeper into this, my branch is here: 

https://github.com/chkal/ozark/tree/bundle

chkal:
Also it looks like that the API classes are now included in the main Ozark archive. That doesn't look correct to me.

ozark/pom.xml
```diff
@@ -8,11 +8,52 @@
...
+        <packages.export>
+            javax.mvc.*,org.glassfish.ozark.*
+        </packages.export>
+    </properties>
+
     <build>
         <plugins>
             <plugin>
+                <groupId>org.glassfish.build</groupId>
+                <artifactId>spec-version-maven-plugin</artifactId>
```
lefloh: Maven warning. Please add `<version>1.2</version>`

ozark/pom.xml
```diff
@@ -79,13 +120,37 @@
...
             </plugin>
+            <!--
+             This plugin is reponsible for packaging artifacts
+             as OSGi bundles.  Please refer to
+             http://felix.apache.org/site/apache-felix-maven-bundle-plugin-bnd.html
+             for more information about how to use this plugin.
+           -->
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
```
lefloh: Maven warning. Please add `<version>3.0.1</version>`.

ozark/src/main/java/org/glassfish/ozark/cdi/OzarkCdiExtension.java
```diff
@@ -124,4 +137,45 @@ public static synchronized void addObservedEvent(Class<? extends MvcEvent> event
...
     }
+
+    /**
+     * Registers given annotated types to be scanned during bean discovery
+     *
+     * @param beforeBean The BeforeBeanDiscovery
+     * @param beanManager The BeanManager
+     */
+    public void register(@Observes BeforeBeanDiscovery beforeBean, BeanManager beanManager) {
+
```
lefloh: It's a little bit confusing having two methods observing `BeforeBeanDiscovery`. Could we add the registration to `#beforeBeanDiscovery`?

lefloh:
@chkal the ozark bundle exports them: 

    <packages.export>
        javax.mvc.*,org.glassfish.ozark.*
    </packages.export>

Don't know much about Glassfish/OSGI but Jersey does the [same here](https://github.com/jersey/jersey/blob/master/bundles/jaxrs-ri/pom.xml#L284) for the JAX-RS API.

ozark/pom.xml
chkal: Good catch. I'll add this.

ozark/pom.xml
chkal: Good catch No. 2. ;)

chkal:
@lefloh Does "exporting" mean, that they end up in the bundle JAR? IMHO there shouldn't be any dependencies "shaded" into the bundle.

rahmanusta:
@chkal @lefloh Much of the impls uses this style (bundles spec + impl as single bundle jar), but we can create an additional gf maven project for only glassfish module.

chkal:
@rahmanusta So you are talking about a special `bundle` Maven module which basically repackages the ordinarily Ozark JAR together with the API into a Glassfish bundle? It looks like Jersey is doing it this way. And IMO this makes sense. This way we have a standard JAR file **and** a Glassfish compatible bundle.

rahmanusta:
@chkal Actually bundle is a nothing more a classical jar, the only difference is some metadata properties in MANIFEST file. In that way, it is a choice, it may not be wanted to bundle spec packages into the bundle.

chkal:
I know that a bundle is basically a JAR file with some special entries in the manifest. But in the case of Ozark changing the packaging to a bundle basically changes the contents of the JAR archive (because it now also contains the API).

Maybe I've a very conservative view on this, but in my opinion the API should be kept separate from the implementation.

BTW: The JAX-RS API specification is not a bundle but a regular JAR file. And it looks to me that Jersey is also a regular JAR file but has a special Maven module which repackages Jersey as a bundle (perhaps also containing the API? I didn't check).

To be honest, I prefer how Jersey handles this. Other thoughts?

@rahmanusta: Are there any other specs you know of which use one or the other way?

rahmanusta:
As we see Jersey uses a bundle directory for only OSGI and other parts are classical osgi. Majority of the impls use this style, but also I can suggest the following style. It meets with the jersey, but be sure that this style is also ok for me.

* spec => only spec api and packaged as bundle, so it is usable in any container.
* impl => only impl codes and packaged as bundle, so it is usable in any container.
* bundle or gf => spec and api codes are together and packaged as bundle, so it is usable in any container.

ozark/pom.xml
rahmanusta: Its version comes from the parent, I haven't seen any bundle that is declared version explicitly on maven-bundle-plugin.

ozark/pom.xml
rahmanusta: It is the same case that I described below :)

ozark/pom.xml
lefloh: The warning is quite clear:
```
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.glassfish.ozark:ozark:bundle:1.0.0-m03-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.felix:maven-bundle-plugin is missing. @ org.glassfish.ozark:ozark:[unknown-version], .../pom.xml, line 129, column 21
[WARNING] 'build.plugins.plugin.version' for org.glassfish.build:spec-version-maven-plugin is missing. @ org.glassfish.ozark:ozark:[unknown-version], .../pom.xml, line 33, column 21
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

rahmanusta: Maven warnings generally messy if you know what to do. Version information comes from the parent "jvnet-parent" so it is not necessary to duplicate it. No one (other specs and impls) worried about that.

lefloh: The version can't come from the parent because there is no version defined in [jvnet-parent](https://repo1.maven.org/maven2/net/java/jvnet-parent/5/jvnet-parent-5.pom). This means that it's unpredictable which version will be used for the build so the warning makes absolutely sense.

chkal: I agree with @lefloh. The version isn't set in the parent, so Maven will always try to use the latest released one. But this makes the build unpredictable. That's what the Maven warning basically is telling us.

Jersey for example sets the version of the plugin in it's own parent pom:

https://github.com/jersey/jersey/blob/master/pom.xml#L558-578

I really think omitting Maven plugin versions is bad practice (unless you inherit them from a parent). So I think we should definitely set the version.

rahmanusta: Sorry @lefloh @chkal  I supposed it comes from the parent without checking it :) Okay, let's set it. Which version you should prefer ? The latest one ?

chkal: Sure, the latest one should be fine if it works without problems.

rahmanusta: Okay, I will be able to send a PR for that next week, but instead of waiting one week, you can make wanted changes before merging the PR.

lefloh: If you haven't defined one the latest one will be used so now we're pretty sure that the latest one is ok. But by defining it we're sure that we'll get the same result in the future if there's a new version from the plugin.

ozark/src/main/java/org/glassfish/ozark/cdi/OzarkCdiExtension.java
rahmanusta: Yes @lefloh , it seems the hook is duplicated. We should merge it.

chkal:
To be honest, I see OSGi as some kind of special case. But this may be just my opinion as I haven't so much experience with OSGi.

I would prefer the style most specs and RIs seems to use which is having a standard JAR for the implementation with an separate maven module which repackages the implementation into an OSGi bundle which may or may not include the API (not sure what works best). 

Of cause this is just my personal opinion. So having other votes for one of the options would be great.

ozark/pom.xml
rahmanusta: Yes, I mean the latest one at this time. If it doesn't get a constant version anywhere, further changes may break the builds.

chkal: @rahmanusta I'll add the versions to my personal branch which includes your patch + minor additions. But please note that we also have to implement the packaging as discussed below and that we still have the failing tests. So perhaps it would make sense to create a new clean PR with the structure discussed below and check whether this gets the tests green?

rahmanusta: @chkal Okay, we agreed on that.

chkal: But no need to hurry. Voxxed Days goes first. :smile:

rahmanusta:
@chkal To make packaging declaration "bundle", it has no any disadvantages, but it has an advantage. For example, if you put the api bundle and impl bundle together into an OSGI container, it will run seamlessly. For example, json-p's api and bundle packages are bundled

https://java.net/projects/jsonp/sources/git/content/api/pom.xml?rev=4712f1041cb3fd6d99ece141a8b0fed39676baaa

https://java.net/projects/jsonp/sources/git/content/impl/pom.xml?rev=4712f1041cb3fd6d99ece141a8b0fed39676baaa

Btw. Not bundling the spec packages into impl bundle has only one advantage, it is just size of the jar bundle. Because, we declared spec-api dependecy as optional in the impl already, if someone adds a different spec version, it will override the default one.

chkal:
Hey @rahmanusta. Thanks for the explanation. My assumption also was that changing the packaging type to bundle just adds some more entries to MANIFEST.MF. But then I saw that after applying your patch the actual content of the resulting JAR file changed as the API was now bundled into the JAR. And I don't think that the main Ozark JAR should include the API classes in the archive.

However, if this is just some kind of configuration of the bundle plugin, then we should change that. In this case I'm also OK with changing the main Ozark module to packaging type bundle. But for me it is very important that the only difference to the classical packaging is that there are some more entries in MANIFEST.MF.

Feel free to fork the branch that I mentioned above. I already added the versions to the plugins. But the most important remaining problem is that we have a ton of test failures.

lefloh:
I would also prefer the Jersey way because we don't know how a container wants to integrate Ozark. As far as I know Glassfish uses for instance a Mojarra bundle including the API classes whereas Wildfly separates both. It's easier for them to choose if we provide a simple jar and a separate bundle.

ozark/pom.xml
rahmanusta: Thank you @chkal for your understanding :) I will look into all details here we talked about Ozark. Thanks

rahmanusta:
Hi @chkal @lefloh ,

I updated my pull request, all tests run on my locale even so Travis failed :) Could you please test and review it?

I also look into details for jersey project structure, and I understand that there are several build structure summurized below;

**jsonp has 2 packages** ()

javax.json:javax.json-api (api) [bundle]
org.glassfish.json:javax.json (api + impl) [bundle]

**jsf has 3 packages**

com.sun.faces:jsf-api (api) [bundle]
com.sun.faces:jsf-impl (impl) [bundle]
org.glassfish:javafx.faces (api + impl) [bundle] [glassfish uses it as module]

**jersey has many packages**

Every packages are seperated [bundle]s, [glassfish uses them as modules]

chkal:
Hey @rahmanusta,

I'll try to find some time in the next days to have a deeper look at this. Unfortunately I'm busy as hell ATM. So it may take some days. Sorry about that. :(

A few first thoughts:

Basically I really like the JSF structure. For me it is important that we end up with having a JAR which contains only the API and a JAR that contains only the RI without the API. And this doesn't seem to be the case in the JSONP packaging. That's why I prefer the JSF structure. We will get individual JAR files AND we get a Glassfish compatible additional module which packages API and RI. Anyone else has some thoughts on this?

Another thing: Could you perhaps explain the other changes you did in the latest version of the pull request:

* You are doing an additional `CDI.current().getBeanManager()` call in `RedirectScopeManager` although there is already a `BeanManager`.
* You replaced the `@Inject @Any Instance<SPI>` code with some manual lookup. Why is this? Did you find out why the view engine lookup was broken in the first version of your patch?
* You changed the the dependency scope of Ozark in the `test` pom.xml? Why? I would expect that this breaks the integration tests because in the integration test there is no Ozark bundle installed in Glassfish and with changing the scope it won't be packed withing the WAR any more.

Also, I'm wondering whether we should start to split the work on this into smaller chunks. The PR is getting very big and difficult to review. It looks like we have these chunks:

  1. Remove beans.xml and register all relevant beans via the CDI extension
  2. Adjust the way that is used to lookup implementations for SPIs (Currently we use `@Inject @Any` which doesn't seem to work)
  3. Perhaps other adjustments like the the change you did in RedirectScopeManager.
  4. Finally modify the pom.xml files to build the bundles

What do you think? We could for example already start pushing the changes for (1) into the master branch because this is something we definitely need to do and which will break nothing at the moment. Working in smaller steps may be a good thing?

Any other thoughts by @lefloh, @ivargrimstad, @spericas, @manfredriem ?

rahmanusta:
Hey @chkal,

No hurry! You listed the PR in 4 items in list, and I agree on that. I will arrange new PRs seperated.

When testing I get that `@Inject @Any Instance<SPI>` doesn't provide added new beans in runtime, I'm not sure but it just injects available beans in initialization phase. So that, I put dynamic lookup code for ViewEngineFinder and LocaleResolverChain bean.

When finding the issue I tried several things and I'm not sure why I used `CDI.current().getBeanManager()` instead of injected "BeanManager". Theoretically it should work with the injected BeanManager, but I will test it again.

I was moving the Glassfish module directory, yes, I will change the provided scope back in test project.

JSF's style is also good to me, it is not hard to move this style. One question is for that, 

javax.mvc:javax.mvc-api (api bundle)
org.glassfish.ozark:ozark (impl bundle)

What should be the artifact and group name for Glassfish bundle? Because it should be shared on Maven repo seperately. Also, keep in mind that if we put api and impl jars into glassfish separately, it also works. Actually we don't need single (api + impl) bundle but we may need to modify something special in the future for Glassfish only.

chkal:
Hey @rahmanusta,

> You listed the PR in 4 items in list, and I agree on that. I will arrange new PRs seperated.

That would be awesome. Perhaps you could start with the removal of beans.xml and the manual registration of CDI beans with the extension? I think we all agree that this is a required step and we can merge it very quickly.

> When testing I get that @Inject @Any Instance<SPI> doesn't provide added new beans in runtime, I'm not sure but it just injects available beans in initialization phase. So that, I put dynamic lookup code for ViewEngineFinder and LocaleResolverChain bean.

Ah, interesting. So you say it may have something to do with the fact that the owning bean is application scoped? However, that's weird, as the other view engines aren't added at runtime but simply are defined in another bean archive.

However, I like the idea of using some utility code to lookup all implementations of an SPI. This way we can modify the lookup code more easily. And there are some edge cases we should support correctly. See for example [this great blog post](https://struberg.wordpress.com/2012/05/11/the-tricky-cdi-serviceprovider-pattern/). 

> When finding the issue I tried several things and I'm not sure why I used CDI.current().getBeanManager() instead of injected "BeanManager". Theoretically it should work with the injected BeanManager, but I will test it again.

Ok, let's try to modify the code as little as possible. 

> I was moving the Glassfish module directory, yes, I will change the provided scope back in test project.

Great!

> Also, keep in mind that if we put api and impl jars into glassfish separately, it also works. Actually we don't need single (api + impl) bundle but we may need to modify something special in the future for Glassfish only.

Ah, OK, I wasn't aware of that. So we won't even need a separate bundle at this point? In this case we should perhaps keep it simple for now and just use the plain API bundle and the Ozark bundle for Glassfish. I agree that there may be the need for more Glassfish integration code later, but currently these is no such code. So I don't see any real benefit of having a separate bundle which contains the API + Ozark in a single file.

lefloh:
Hey @rahmanusta,

@chkal is right. The test-apps can't be deployed any more because of the changed scope (`java.lang.ClassNotFoundException: javax.mvc.Models`) and so the tests are failing. Maybe it's working on your local Glassfish because you installed MVC as bundle?

Making smaller steps sounds like a good idea!

rahmanusta:
Hey @lefloh, you are right. I will do the new PRs.

rahmanusta:
Closed, will continue a clean PR.

PR #5, opened 2016-04-23
OZARK-65: Log an error if a controller is not CDI managed only

lefloh:
Fix for [OZARK-65](https://java.net/jira/browse/OZARK-65).

### DIFF
--- a/ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
+++ b/ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
@@ -42,16 +42,20 @@ package org.glassfish.ozark.jersey;
 import org.glassfish.jersey.internal.spi.AutoDiscoverable;
 import org.glassfish.jersey.internal.spi.ForcedAutoDiscoverable;
 import org.glassfish.ozark.MvcContextImpl;
+import org.glassfish.ozark.binding.BindingInterceptorImpl;
 import org.glassfish.ozark.core.ViewRequestFilter;
 import org.glassfish.ozark.core.ViewResponseFilter;
 import org.glassfish.ozark.core.ViewableWriter;
 import org.glassfish.ozark.security.CsrfProtectFilter;
 import org.glassfish.ozark.security.CsrfValidateInterceptor;
-import org.glassfish.ozark.binding.BindingInterceptorImpl;
+import org.glassfish.ozark.util.AnnotationUtils;
 
 import javax.annotation.Priority;
+import javax.ejb.Stateful;
+import javax.ejb.Stateless;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.CDI;
+import javax.faces.bean.ManagedBean;
 import javax.mvc.annotation.Controller;
 import javax.servlet.ServletContext;
 import javax.ws.rs.ConstrainedTo;
@@ -59,7 +63,13 @@ import javax.ws.rs.RuntimeType;
 import javax.ws.rs.core.Configuration;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.FeatureContext;
+import java.lang.annotation.Annotation;
 import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.glassfish.ozark.util.AnnotationUtils.getAnnotation;
 import static org.glassfish.ozark.util.CdiUtils.newBean;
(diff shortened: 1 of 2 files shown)


ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
```diff
@@ -87,27 +101,47 @@ public void configure(FeatureContext context) {
...
-            context.register(CsrfProtectFilter.class);
-
-            // Initialize application config object in Mvc class
-            final BeanManager bm = CDI.current().getBeanManager();
-            final MvcContextImpl mvc = (MvcContextImpl) newBean(bm, MvcContextImpl.class);
-            mvc.setConfig(config);
-            mvc.setContextPath(servletContext.getContextPath());
+        Set<Class<?>> controllers = Stream.concat(config.getClasses().stream(), config.getInstances().stream().map(o -> o.getClass()))
+                .filter(this::isController).collect(Collectors.toSet());
+        if (controllers.isEmpty() || hasUnsupportedTypes(controllers)) {
```
chkal: To be honest, I'm not completely sure if we should abort the initialization in case of a single problematic controller. Wouldn't it make more sense to just log a warning and continue?

ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
```diff
@@ -87,27 +101,47 @@ public void configure(FeatureContext context) {
...
+    boolean isController(Class<?> c) {
         return getAnnotation(c, Controller.class) != null ||
                 Arrays.asList(c.getMethods()).stream().anyMatch(m -> getAnnotation(m, Controller.class) != null);
     }
+
+    /**
+     * <p>MVC classes are required to be CDI-managed beans only.
+     * Managed Beans or EJBs are not allowed (MCV-Spec ch. 2.1.1).</p>
+     */
+    boolean hasUnsupportedTypes(Set<Class<?>> controllers) {
```
chkal: What about changing the signature to:

```java
boolean hasUnsupportedTypes(Class<?> controller)
```

This way the method would be simpler to test. And you could check if there are problematic controllers and log their names simply by using something like:

```java
controllers.stream()
    .filter( c -> hasUnsupportedTypes(c) )
    .forEach( c -> LOG.server( "...." + c.getSimpleName() ) );
```

Of cause this would create one log entry for each controller, but that's IMO even better compared to the comma separated list of class names.

ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
lefloh: On the other hand it will lead to errors which are hard to find if we continue (like a NPE in an Ozark Utils class). But I'm also not sure about that. Other opinions?

ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
lefloh: Good idea! I'll update the PR.

chkal: Great! :+1:

ozark/src/main/java/org/glassfish/ozark/jersey/OzarkFeature.java
chkal: Yeah, the best way would actually be to fail with a runtime exception. But as you wrote in the comments this isn't possible as Jersey swallows all exceptions.

So it looks like we have two options:

1. Log an error message and abort the Ozark bootstrapping process. The application will continue to start up but all MVC functionality will simply not work. My fear is that people may overlook the error message and then wonder why nothing works as expected.
2. Log an warning message and continue to bootstrap Ozark. The application will continue to start up and MVC will basically work, but problematic controllers may fail with weird NPEs. Not so sure if this is much better than the first choice. In this case people may also overlook the warning.

I guess both ways are not very satisfying. We could do some research and try to find out why Jersey swallows the exception. Because this would be actually the best solution:

* Fail with an runtime exception which will break the application startup. The developer will immediately see the error and then knows how to fix it.

lefloh: Another way could be check for unsupported controllers in the `OzarkContainerInitializer`. Here we could stop the deployment but would need to search for all resource classes on our own and would maybe find others then Jersey...
I opened [JERSEY-3101](https://java.net/jira/browse/JERSEY-3101). Maybe they could at least log an Exception with a higher level then `FINE`.

chkal:
Updated issue reference: #81

chkal:
There have been quite some changes in these classes. That's why the PR doesn't merge anymore.

Also, I think we will need a CDI extension in the near future which collects metadata about controllers. This would be a great place for this check.

Therefore, I would like to close this for now. But we will keep #81 to address it later on.

Any objections?

chkal:
Closing due to the lack of activity on this one.

PR #6, opened 2016-04-23
OZARK-73: Return the correct status code when using @View 

lefloh:
Return the correct status code when using @View and throwing an Exception.

Fix for [OZARK-73](https://java.net/jira/browse/OZARK-73).


chkal:
Looks good to me! I like that the actual change to `ViewResponseFilter` is pretty minimal. And the test coverage is awesome! Thank you. :beers:

chkal:
Merged! Thanks a lot! :+1:

PR #11, opened 2016-05-17
Register Ozark beans with CDI extension mechanism

rahmanusta:
This is the first part of my work "Register Ozark beans with CDI extension mechanism"

### DIFF
--- a/ozark/src/main/java/org/glassfish/ozark/cdi/OzarkCdiExtension.java
+++ b/ozark/src/main/java/org/glassfish/ozark/cdi/OzarkCdiExtension.java
@@ -39,6 +39,24 @@
  */
 package org.glassfish.ozark.cdi;
 
+import org.glassfish.ozark.MvcAppConfig;
+import org.glassfish.ozark.MvcContextImpl;
+import org.glassfish.ozark.binding.BindingInterceptorImpl;
+import org.glassfish.ozark.binding.BindingResultImpl;
+import org.glassfish.ozark.binding.ConstraintViolationTranslator;
+import org.glassfish.ozark.core.*;
+import org.glassfish.ozark.engine.FaceletsViewEngine;
+import org.glassfish.ozark.engine.JspViewEngine;
+import org.glassfish.ozark.engine.ViewEngineFinder;
+import org.glassfish.ozark.event.*;
+import org.glassfish.ozark.locale.DefaultLocaleResolver;
+import org.glassfish.ozark.locale.LocaleRequestFilter;
+import org.glassfish.ozark.locale.LocaleResolverChain;
+import org.glassfish.ozark.security.CsrfImpl;
+import org.glassfish.ozark.security.CsrfProtectFilter;
+import org.glassfish.ozark.security.CsrfValidateInterceptor;
+import org.glassfish.ozark.security.EncodersImpl;
+import org.glassfish.ozark.util.CdiUtils;
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.BeanManager;
@@ -72,6 +90,56 @@ public class OzarkCdiExtension implements Extension {
      */
     public void beforeBeanDiscovery(@Observes final BeforeBeanDiscovery event, BeanManager beanManager) {
         event.addScope(RedirectScoped.class, true, true);
+
+        CdiUtils.addAnnotatedTypes(event, beanManager,
+
+                // .
+                MvcAppConfig.class,
+                MvcContextImpl.class,
+
+                // binding
+                BindingResultImpl.class,
+                BindingInterceptorImpl.class,
+                ConstraintViolationTranslator.class,
+
+                // core
+                Messages.class,
+                ModelsImpl.class,
+                ViewableWriter.class,
+                ViewRequestFilter.class,
+                ViewResponseFilter.class,
+
+                // engine
+                FaceletsViewEngine.class,
+                JspViewEngine.class,
+                ViewEngineFinder.class,
+
+                // security
+                CsrfImpl.class,
+                CsrfProtectFilter.class,
+                CsrfValidateInterceptor.class,
+                EncodersImpl.class,
+
+                // util
+                CdiUtils.class,
+
+                // cdi
+                RedirectScopeManager.class,
+
+                //event
+                AfterControllerEventImpl.class,
+                AfterProcessViewEventImpl.class,
+                BeforeControllerEventImpl.class,
+                BeforeProcessViewEventImpl.class,
+                ControllerRedirectEventImpl.class,
+                MvcEventImpl.class,
+
+                //locale
+                LocaleRequestFilter.class,
+                LocaleResolverChain.class,
+                DefaultLocaleResolver.class
+
+        );
     }
 
     /**
--- a/ozark/src/main/java/org/glassfish/ozark/util/CdiUtils.java
+++ b/ozark/src/main/java/org/glassfish/ozark/util/CdiUtils.java
@@ -43,6 +43,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.context.spi.CreationalContext;
 import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.BeforeBeanDiscovery;
 import javax.inject.Inject;
 import java.util.Set;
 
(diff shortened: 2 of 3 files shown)


rahmanusta:
I wonder why it is failed :) @chkal @lefloh

rahmanusta:
It is failed because we should update the ViewEngineFinder, so you can merge it, it will be fixed in upcoming PR

rahmanusta:
Interestingly injected BeanManager doesn't find registered beans, but CDI.current does that. @chkal @lefloh May it be a CDI bug ?

chkal:
Hey @rahmanusta. Feel free to include the ViewEngineFinder modification required in this PR. Let's try to keep the tests green.

But actually it is very weird that `@Inject BeanManager` doesn't work but `CDI.current().getBeanManager` does. Actually this should resolve to the same BeanManager.

chkal:
I can confirm for some reasons there are two bean managers:

Calling `toString()` on the result of `@Inject BeanManager` returns:

    Weld BeanManager for org.glassfish.ozark.cdi.OzarkCdiExtension [bean count=57]

And doing the same for `CDI.current.getBeanManager()` returns:

    Weld BeanManager for root_test-freemarker [bean count=28]

Now that's confusing.... :confused:

rahmanusta:
@chkal This is the answer, at least we know there can be more different BeanManager for each module. I found this http://stackoverflow.com/questions/22807435/understanding-cdi-weld-in-multi-module-application

chkal:
Thanks, I'll have a depper look at the second link which looks promising...

My findings so far. Before applying your patch we have this:

```
@Inject BeanManager:
    Weld BeanManager for WEB-INF/lib/ozark-1.0.0-m03-SNAPSHOT [bean count=67]]]
CDI.current().getBeanManager():
    Weld BeanManager for WEB-INF/lib/ozark-1.0.0-m03-SNAPSHOT [bean count=67]]]
```

After applying it we have:

```
@Inject BeanManager:
    Weld BeanManager for org.glassfish.ozark.cdi.OzarkCdiExtension [bean count=57]]]
CDI.current().getBeanManager():
    Weld BeanManager for root_test-freemarker [bean count=28]]]
JNDI lookup:
    Weld BeanManager for test-freemarker [bean count=30]]]
```

rahmanusta:
Very interesting, how we will go forward? Do you still have this opinion "**Feel free to include the ViewEngineFinder modification required in this PR. Let's try to keep the tests green.**"

chkal:
I'll try to contact Martin Kouba regarding this and see what he thinks. I guess the Weld project lead can help us here.

I think doing the ViewEngineFinder modification you did in the other PR will work, but it would be great to get some clarification about the situation because we are using `@Inject BeanManager` also at other places...

rahmanusta:
Let's try to join them this conversation,

Hi @antoinesd, @mkouba , we have an issue about accessing BeanManager. When we apply a CDI extension to Ozark project, we are getting 3 different BeanManager that Christian described it above. It is very complicated for us, How should we move ?

Thanks.

rahmanusta:
> BTW: Only @Inject BeanManager (after applying your patch) doesn't find the additional view engines.

Yes, `CDI.current().getBeanManager()` and `JNDI lookup for java:comp/BeanManager:` work for us,  but we should clarify this issue.

mkouba:
Hi guys, in Weld there is not just a single `BeanManager`. Instead we have a `BeanManager` per bean archive. 

I'm not really sure what the original motivation was but I suppose it's mainly because some BM methods require some context - e.g. `getBeans()` _"Returns the set of beans which have the given required type and qualifiers and **are available for injection in the module or library** containing the class into which the BeanManager was injected..."_. 

Note that the spec does not say anything like "there is a single BeanManager per application".

rahmanusta:
Thank you @mkouba, it is strange to me but it is good to clarify current state of BeanManager from our side.

@chkal I echoed bean instances in a sheet to compare our situation clearly. Please look at that. https://docs.google.com/spreadsheets/d/1LHjq7Ph9rJlW3N7AFMQVOVy0nBIEAVwrXuBDz6rnr4k/edit?usp=sharing

chkal:
Hey @mkouba,

let me try to explain what we are doing here. Currently Ozark is a standard CDI archive with a `beans.xml` file. We have a test web application which basically looks like this:


```
test-freemarker.war
├── ...
└── WEB-INF
    ├── beans.xml
    ├── classes
    │   └── ...
    └── lib
        ├── ....
        ├── ozark-1.0.0-m03-SNAPSHOT.jar
        └── ozark-freemarker-1.0.0-m03-SNAPSHOT.jar
```

Now if we try to obtain the bean manager via different methods we are getting this result:

```
@Inject BeanManager into Ozark class:
    Weld BeanManager for WEB-INF/lib/ozark-1.0.0-m03-SNAPSHOT [bean count=67]]]

CDI.current().getBeanManager() from an Ozark bean:
    Weld BeanManager for WEB-INF/lib/ozark-1.0.0-m03-SNAPSHOT [bean count=67]]]

JNDI lookup for java:comp/BeanManager:
    Weld BeanManager for test-freemarker [bean count=30]]]
```

This basically looks fine to me. Ozark gets its BM and the web application BM is visible via JNDI. Everything fine so far.

Now we are in the process of preparing Ozark do be deployable as a Glassfish module. According to our knowledge this means that we cannot use `beans.xml` in the main Ozark JAR any more. So this pull request removes `beans.xml` from the main Ozark JAR and adds some code to `OzarkCdiExtension` which observers `BeforeBeanDiscovery` to register all annotated types required by Ozark.

Now we are getting this when deploying the same application (with Ozark still being in `/WEB-INF/lib`):

```
@Inject BeanManager into Ozark class:
    Weld BeanManager for org.glassfish.ozark.cdi.OzarkCdiExtension [bean count=57]]]

CDI.current().getBeanManager() from an Ozark bean:
    Weld BeanManager for root_test-freemarker [bean count=28]]]

JNDI lookup for java:comp/BeanManager:
    Weld BeanManager for test-freemarker [bean count=30]]]
```

This looks a bit surprising to us. But the main problem is that the BM obtained with the first method (injecting via `@Inject`) doesn't seem to have access to all the beans in other archives. In our case we have an interface `ViewEngine` and there are multiple implementations in different locations.

* Main Ozark JAR
  * `JspViewEngine`
  * `FaceletsViewEngine`
* Ozark Freemarker JAR
  * `FreemarkerViewEngine`

So if we use the injected BeanManager from within Ozark, we don't get the `FreemarkerViewEngine`. In all other cases we get all the view engines. That's a bit confusing. ;)

mkouba:
What does it actually mean "deployable as a Glassfish module"? If the goal is to make it a part of Glassfish then you should not include Ozark jar in the deployment...

chkal:
Actually we want to be able to do both. ;)

1. Install Ozark as a Glassfish module so that Ozark is provided by the container. The WAR file doesn't have to include Ozark in this case any more.
2. For other containers like Tomcat it should be possible to include Ozark in `/WEB-INF/lib`.

Our goal was to build Ozark in way that allows both usage scenarios.

mkouba:
I see. So the second result is Ozark (with extended OzarkCdiExtension) deployed to Tomcat + Weld Servlet? If so, the BM obtained with the first method should see all the beans. However, note that alternatives, interceptors and decorators enabled/selected for a bean archive (not by means of `@Priority`) might be problematic.

chkal:
No, not exactly. Both examples are deployed to latest Glassfish nightly with Weld being provided by Glassfish. The WAR archive contains just Ozark in `/WEB-INF/lib/`.

The only difference between the two examples:

  * Example 1:
    * Ozark JAR contains `beans.xml` with bean-discovery-mode="all"
  * Example 2:
    * Ozark JAR contain **no** `beans.xml` and all relevant CDI beans are registered via an extension.

mkouba:
Ok. The only problematic part I can see is that GlassFish must create a kind of additional/synthetic bean archive for Ozark/OzarkCdiExtension. And this bean archive should follow the Java EE class accessibility requirements. So in this case Ozark should see/be visible from all the bean archives from the webapp.

Also note that this is not clearly defined by the spec, see also https://issues.jboss.org/browse/CDI-233.

chkal:
@mkouba Thanks a lot for sharing your thoughts on this. So do you think this is more a Glassfish than a Weld issue?

smillidge:
@Pandrex247 (in the Payara team) is approaching this by building a separate Payara module to do the bean registration.

rahmanusta:
@smillidge Is there anything different between a GF module and a Payara module ?

arjantijms:
It looks like the three different ways to obtain beans are via:

1. ``@Inject``
1. ``CDI.current().select()`` / ``CDI.current().getBeanManager()``
1. ``initialContext.lookup("java:comp/BeanManager")``

Then there are two locations to put spec implementation jars (e.g. Ozark, Soteria, Mojarra):

1. WEB-INF/lib
1. Inside the server (e.g. /modules in Payara)

And then we have many different servers:

1. Payara
1. JBoss
1. TomEE
1. Liberty
1. etc

From initial testing I found quite a number of those combinations behave differently.

E.g. in Payara ``@Inject`` doesn't work in a container jar, but ``CDI.current()`` and ``initialContext.lookup("java:comp/BeanManager")`` do.

However, in JBoss ``CDI.current()`` does not work in a container jar, but ``initialContext.lookup("java:comp/BeanManager")`` does.

I'll soon try if I can put together a kind of matrix to make it clear what the exact differences are.

rahmanusta:
Thank you @arjantijms for explaining your experience in similar issue. If the JNDI lookup method works for all platforms, we can use it with a simple helper method in our CdiUtils class. What is your opinion @chkal ?

rahmanusta:
My last PR is here https://github.com/mvc-spec/ozark/pull/12, at the end it seems me not too complex and I made it in a single PR.

chkal:
@arjantijms Thanks a lot for sharing your experiences. Some kind of matrix would be really helpful. And I also think that we may need to move the discussion to cdi-dev!?!

PR #12, opened 2016-05-18
Package Ozark as OSGI bundle

rahmanusta:


lefloh:
Thanks for your work again @rahmanusta! One question: Could we remove the injected BeanManger from [RedirectScopeManager](https://github.com/rahmanusta/ozark-1/blob/525bf06f31e119a23ded81091da4484111f721ce/ozark/src/main/java/org/glassfish/ozark/cdi/RedirectScopeManager.java#L93) as this one is not used any more?

rahmanusta:
Thank you @lefloh ! Sure, we can remove it because the injected BM is not used anymore. I uptated the PR now. Thanks.

chkal:
Great! The tests are green! That good!

I'll have a deeper look at it later today. Thanks. :beers:

rahmanusta:
Thank you @chkal !! :) :) :)

chkal:
Hey @rahmanusta,

I finally found some time to have a deeper look at this. Everything fine so far. I've just two remaining questions:

* You changed Ozark's dependency on the spec JAR to be optional? Why is that? Is it to prevent the OSGi bundle plugin to include the API in the bundle? To be honest I don't think this is a good idea because it is in fact NOT optional but required for non OSGi apps. Is there any other way to prevent shading of the API classes into the bundle?
* You removed the injected BeanManager in `RedirectScopeManager` and replaced it with `CDI.current().getBeanManager()`. At first I was not sure why you did that. But the code actually looks up application beans, so we cannot use the injected `BeanManager` for the same reasons as in the other cases, correct?

Thanks for your work! :beers:

chkal:
I just had a look whether having the api as an optional dependency or not has any influence on the content of the resulting JAR bundle. Looks like in both cases the API classes are NOT included, which is exactly like I think it is correct.

So if you don't mind, I would revert changing the dependency to optional before merging your patch.

rahmanusta:
Hi @chkal,

> You changed Ozark's dependency on the spec JAR to be optional? Why is that? Is it to prevent the OSGi bundle plugin to include the API in the bundle? To be honest I don't think this is a good idea because it is in fact NOT optional but required for non OSGi apps. Is there any other way to prevent shading of the API classes into the bundle?

Actually making it optional doesn't make any bad or good effect. Because we are including org.glassfish.ozark* packages into bundle output. We can remove or keep it.

> You removed the injected BeanManager in RedirectScopeManager and replaced it with CDI.current().getBeanManager(). At first I was not sure why you did that. But the code actually looks up application beans, so we cannot use the injected BeanManager for the same reasons as in the other cases, correct?

I think it is the same issue that we saw before, getting bean from injected BeanManager returns null https://github.com/rahmanusta/ozark-1/blob/bundle/ozark/src/main/java/org/glassfish/ozark/cdi/RedirectScopeManager.java#L257 , for that reason I switched the code to use CDI.current().getBeanManager()

Thank you!

chkal:
> Actually making it optional doesn't make any bad or good effect. Because we are including org.glassfish.ozark* packages into bundle output. We can remove or keep it.

Ok, thanks. In this case I don't think that we should make it optional, because making it optional will break many applications which assume that the API is a transitiv dependency.

> I think it is the same issue that we saw before, getting bean from injected BeanManager returns null https://github.com/rahmanusta/ozark-1/blob/bundle/ozark/src/main/java/org/glassfish/ozark/cdi/RedirectScopeManager.java#L257 , for that reason I switched the code to use CDI.current().getBeanManager()

That was my assumption too. So that's fine.

Ok, great. I'll do the merge later today or tomorrow. Many thanks. :tada:

rahmanusta:
Thank you @chkal :+1: :+1: :+1: 

Sould I remove the optional declaration? or will you do it?

Cheers :tada:

chkal:
I'll do it during the merge. Thanks.

chkal:
I just merged the pull request locally and pushed the changes upstream. See https://github.com/mvc-spec/ozark/commit/51255d8537a3a1d917e098a2d4146e110850175d

@rahmanusta: Thanks a lot for your hard work on this! :beers: :tada: :balloon:

rahmanusta:
Thank you so much @chkal :tada: :tada: :tada:

PR #13, opened 2016-05-29
Jersey upgrade to 2.23 and removing of workaround for JERSEY-3069

lefloh:


lefloh:
Seems like the Jersey update breaks the ITs where Ozark is deployed as module. The other ones are green. 

```
Caused by: java.util.ServiceConfigurationError: javax.servlet.ServletContainerInitializer: 
Provider org.glassfish.ozark.servlet.OzarkContainerInitializer not found
    at java.util.ServiceLoader.fail(ServiceLoader.java:239)
    at java.util.ServiceLoader.access$300(ServiceLoader.java:185)
    at java.util.ServiceLoader$LazyIterator.nextService(ServiceLoader.java:372)
```

Will have a look at this later.

lefloh:
The Jersey OSGI module won’t be resolved if we upgrade the minor version. Closing this issue. Might become an option again if we switch to Glassfish 5.

PR #106, opened 2017-06-10
Use latest promoted Glassfish 5 build for integration tests

chkal:
I updated the Glassfish version to use for the integration tests to the latest promoted build of Glassfish 5. Looking forward to see whether the integration tests still pass. ;-)


chkal:
All the tests pass. So I'll merge this one!

PR #107, opened 2017-06-11
Removed VariantSelector and always use "text/html" when @View is used

chkal:
Pull request relates to this discussion on the mailing list:

https://groups.google.com/forum/#!topic/ozark-dev/7cWqXkERs-g

As mentioned in the discussion the `VariantSelector` is only used if a method uses `@View` and returns `null`. In this case JAX-RS doesn't seem to calculate the resulting content type correctly for some reason. Therefore Ozark tries to do this using the `VariantSelector` and falls back to `text/html` if this fails. 

The code is now much simpler and always uses `text/html`. IMO this totally makes sense for Ozark. MVC is a framework for webapps. I've never seen anyone using a different content type than `text/html`. If this ever becomes an issue for someone, we can discuss the individual case and find a solution that works on plain JAX-RS.

This change allows use to remove `VariantSelector` which depends on various Jersey classes.

Let me know if there are any objections.


Daniel-Dos:
I agree with you @chkal , hardly anyone will use anything other than html, I believe this is more suitable for Rest applications.

chkal:
Closing this one without merging as some tests broke. Also #110 provides a much better solution.

Issue #66, opened 2015-06-03
Document support for Redirect


mvcbot:
**Original issue OZARK-48 created by Santiago Pericas-Geertsen:**

Need to document forms of redirect.

PR #109, opened 2017-06-11
Replaced Jersey-dependent BindingInterceptorImpl by CDI ValidationInterceptor

dmaidaniuk:
Pull request relates to this discussion on the mailing list:

https://groups.google.com/forum/#!topic/ozark-dev/pYw5u8VacnA

This PR will allow drop dependency to `org.glassfish.jersey.server.spi.ValidationInterceptor` and apply validation of resources in vendor-independent manner basing on CDI interceptor.


chkal:
Hey @dmaidaniuk. Thanks a lot. I'll try to find some time tomorrow to have a deeper look at the PR.

But it looks like some validation related tests broke. Could you have a look?

dmaidaniuk:
@chkal Yes, I saw it. ValidationIT test failed. I will look into it tomorrow and try to fix problem.

dmaidaniuk:
@chkal It looks like in Jersey exists a bug with processing `@ValidateOnExecution` annotation properly. In `FormController` used for `org.glassfish.ozark.test.validation.ValidationIT` we have next code snippet:
```
    @POST
    @Controller
    @ValidateOnExecution(type = ExecutableType.NONE)
    public Response formPost(@Valid @BeanParam FormDataBean form) {
        if (br.isFailed()) {
            ValidationError validationError = br.getAllValidationErrors().iterator().next();
            final ConstraintViolation<?> cv = validationError.getViolation();
            final String property = cv.getPropertyPath().toString();
            error.setProperty(property.substring(property.lastIndexOf('.') + 1));
            error.setValue(cv.getInvalidValue());
            error.setMessage(cv.getMessage());
            error.setParam(validationError.getParamName());
            return Response.status(BAD_REQUEST).entity("error.jsp").build();
        }
        return Response.status(OK).entity("data.jsp").build();
    }
```
Thus according to [Bean Validation 1.1 documentation](http://beanvalidation.org/1.1/spec/#integration-general-executable) given method shouldn't be validated by Jersey's internal validation mechanism. Same statement present also in [Jersey's own documentation](https://jersey.github.io/documentation/latest/user-guide.html#bv.ValidateOnExecution).

However annotation `@ValidateOnExecution(type = ExecutableType.NONE)` ignored by Jersey and HTTP code 404 returned to client in case with not valid data set in form, which is wrong. I will double-check my assumption and register bug-report for this in Jersey bug-tracker.

Meanwhile I will search for workaround to catch this bug and fix it in Ozark.

*Note:* I also verified given case with other JAX-RS implementation - RESTEasy. And it works properly by just skipping validation stage in such situation.

dmaidaniuk:
I registered bug-report with reproducer-project for mentioned issue in Jersey bug-tracker there:
https://github.com/jersey/jersey/issues/3582
Please vote for it.

chkal:
@dmaidaniuk Hmmm. That's weird. But I don't fully understand why the test worked fine before.

dmaidaniuk:
@chkal It is not weird but obvious. Current implementation of Ozark is trying to catch validation exceptions using `org.glassfish.jersey.server.spi.ValidationInterceptor` interface in case when annotation `@ValidateOnExecution(type = ExecutableType.NONE)` present for method and thus standard validation processing should be disabled. According to MVC spec. in chapter _3.2 Validation Exceptions_ :

>In other words, controller methods can get called even if parameter validation fails as long as they declare themselves capable of handling errors.
> A controller class that, either directly or indirectly via inheritance, defines a field or a property of type
javax.mvc.binding.BindingResult will have its controller methods called even if an error is encountered
while validating parameters.

That's why in given example for this chapter added line `@ValidateOnExecution(type = ExecutableType.NONE)` which turn-off standard validation flow for JAX-RS 2.0 implementations for compatibility reasons, I guess:

```
@Controller
@Path("form")
public class FormController {

  @Inject
  private BindingResult br;

  @Inject
  private ErrorDataBean error;

  @POST
  @ValidateOnExecution(type = ExecutableType.NONE)
  public Response formPost(@Valid @BeanParam FormDataBean form) {
    if (br.isFailed()) {
      // fill out ErrorDataBean ...
      return Response.status(BAD_REQUEST).entity("error.jsp").build();
    }
    return Response.status(OK).entity("data.jsp").build();
  }
}
```
It means that MVC implementations should implement own way for catching and processing validation errors in called method instead of trying to override implementation-specific JAX-RS validation mechanism.

Daniel-Dos:
Interesting your observation.

For me to enter, the validation occurs, however with error. So instead of presenting an error, it lets it go as if nothing has happened?

In the case of the validation mechanism, instead of using a ready solution, we should have the one specific to the MVC and that works correctly, would it?

chkal:
@dmaidaniuk Thanks for the explanation. I also had a look the old interceptor code and now I think I fully understand it. Basically the interceptor will NOT rethrow `ContraintViolationException`s if they are added to `BindingResult`. Therefore JAX-RS never sees them. Correct?

The interesting thing here is that with this implementation `@ValidateOnExecution(NONE)` is actually not really required. One the one hand that is good news because we were planning to get rid of the requirement to add it anyway (see #66). On the other hand we will lose this with the new implementation because without adding `@ValidateOnExecution(NONE)` JAX-RS will validate the request again.

Hmmm. However, for now we can at least confirm that this is a Jersey issue. I wonder whether this issue is a regression introduced in one of the latest releases. Because I'm pretty sure that it worked some time ago. The validation integration was rewritten a few times and back then we definitely had to add the annotation to get it to work correctly.

dmaidaniuk:
@chkal  It seems that GitHub linked the wrong issue mentioned by you. Correct one placed in MVC spec space there: https://github.com/mvc-spec/mvc-spec/issues/66  
I think we can accept this enhancement №66 with removal `@ValidateOnExecution(type = ExecutableType.NONE)` part and adding instead of it some new annotation, e.g. proposed `@MvcValidated`. Or we even can just drop BV validation annotation and don't add anything instead of it. It is clearly defined in MVC spec when implementations should care about a special way of MVC validation: 

> A controller class that, either directly or indirectly via inheritance, defines a field or a property of type
javax.mvc.binding.BindingResult will have its controller methods called even if an error is encountered
while validating parameters.

And `@ValidateOnExecution` only mentioned in an example and a footnote for related article. Thus those 2 things should be removed from MVC spec too.
I guess we shouldn't force implementors to rely on this newly introduced annotation and leave a space for different solutions here. 
For Ozark we can just annotate programmatically particular methods by original annotation from BV. It can be done in the same way I do now for this new CDI interceptor with special technical annotation [MvcValidation](https://github.com/mvc-spec/ozark/pull/109/commits/34696f277a69a1028ec843d79e5106ec95c9ef39#diff-9b8e1fe57a4e0f07e48592554caa03f6). Thus we will not break or hack standard JAX-RS validation flow but just turn-off it in a programmatic way rather than current declarative one.
Does it make sense?

chkal:
@dmaidaniuk:

Thanks for correcting the issue link. I forgot to include the repository name so it was broken.

Regarding `@ValidateOnExecution`: I always disliked that users had to put `@ValidateOnExecution(NONE)` on their controllers. That definitely was the case in the first versions of the spec. So I'm actually very surprised that currently `@ValidateOnExecution(NONE)` is actually not required any more. And if so, we should of cause remove it from all examples and the spec itself.

However, for me it looks like that `@ValidateOnExecution` is required with your new implementation. Although it currently doesn't work correctly in Jersey. I don't think that I understand you idea regarding a "technical annotation" to "annotate programmatically". With "technical annotation" you are referring to an interceptor binding annotation? But how could we annotate resource methods programmatically with `ValidateOnExecution()`. Could you elaborate on this a bit more?

dmaidaniuk:
@chkal , please look to this blog post ["How to add Annotations at Runtime to a Java class method using Javassist? (Part 1 ) "](http://ayoubelabbassi.blogspot.com/2011/01/how-to-add-annotations-at-runtime-to.html). There described a more clear idea which I mentioned in my previous post. Using `Javassist` library we can add to particular classes new annotation(s) on runtime. In our case, we are interested in adding `@ValidateOnExecution(NONE)` annotation programmatically for Controller-classes with injected `BindingResult` field or property. Such action will deactivate JAX-RS standard validation flow and don't force developers to mark those methods by thyself. This obvious job will be done by Ozark on a background.

chkal:
To be honest, I don't think that this will work. Application servers like Glassfish and Wildfly typically do the annotation scanning once before the app is deployed. So even if we try to add annotations dynamically, we would have to do it before the annotation scanning happens. But it is worth a try if you like.

chkal:
BTW: Doesn't `CtClass.toClass()` basically create a new class instead of modifying the existing one?

dmaidaniuk:
@chkal I'm not so familiar with `Javassist` library and can't be 100% how it works. From what I see that this library already used by Jersey's HK2 injection framework and thus included in Ozark dependencies:

> [INFO] |  +- org.glassfish.hk2:hk2-locator:jar:2.4.0-b31:provided
> [INFO] |  |  \- org.javassist:javassist:jar:3.18.1-GA:provided

I go in a different way proposed on [this SO answer](https://stackoverflow.com/a/30287201/2881869) where used pure Java Reflection API for Java 8. And I made required changes for my PR in [next commit](https://github.com/dmaidaniuk/ozark/commit/41bdf855db57768fb3121b4e5e045be47c685675). I tested those changes with next Controller resource:
```
@Controller
@Path("validation-test")
public class ValitationTestController {

    @Inject
    private BindingResult br;

    @Inject
    private Models models;

    @GET
    public String sayHello(@QueryParam("name") @Size(min = 1) @Pattern(regexp = "[A-Z][a-zA-Z]*") String name) {
        if (br.isFailed()) {
            this.models.put("bindingResult", br);
            return "/WEB-INF/jsp/error.jsp";
        }
        this.models.put("text", "Hello " + name);
        return "/WEB-INF/jsp/validation.jsp";
    }

}
```
The approach works fine for `RESTEasy/Wildfly 10` and doesn't work with `Jersey/Payara 4` due to registered by me issue with [broken validation handling](https://github.com/jersey/jersey/issues/3582).
Proposed solution is only POC and thus should be improved for sure. But it works for main application servers and JAX-RS implementations.

chkal:
Ok, interesting. Perhaps an approach we could consider.

Too bad this Jersey bug is blocking us now. I just tried Jersey 2.25.1 in Tomcat and the bug doesn't occur there. Which Jersey version and environment did you test?

dmaidaniuk:
@chkal For testing I used fresh Payara 4.1.2.172 server which based on GlassFish 4.1.2. Also I tested my reproducer with promoted GlassFish 5 where it is also present. I see that `Bundle-Version: 2.22.2` is for Jersey libraries used in Payara 4.1. I used [next post](https://blog.dejavu.sk/2014/01/21/updating-jersey-2-in-glassfish-4/) to get this info as it is not so easy to find it somewhere in GlassFish release notes or documentation.

chkal:
Thanks. So it looks like this is broken for quite some time now.

I posted a question regarding the use of `@ValidateOnExecution` to the JAX-RS spec list:

https://javaee.groups.io/g/jaxrs-spec/message/187

dmaidaniuk:
Reported blocker bug jersey/jersey#3582 solved now. Let's see in which Jersey release it will be included. Meanwhile, my idea with CDI validation interceptor is unblocked now and can be taken as default implementation, I guess. @chkal , what do you think?

Daniel-Dos:
Hi @dmaidaniuk  The correction is in this version of Jersey: [2.26-b07](https://github.com/jersey/jersey/releases/tag/2.26-b07) .
 
{  },s

dmaidaniuk:
Hi @Daniel-Dos,

Thanks for info! Well, we can upgrade Glassfish 5 Jersey module by ourself on Travis, I assume. ;) It should be quite easy to do as it is micro-version upgrade.

Regards,
Dmytro

Daniel-Dos:
Hi Dmytro,

There in folder bin,  a script of update for jersey.

But I do not know how it works on Travis.

https://github.com/mvc-spec/ozark/blob/master/bin/update-jersey-in-gf.sh
{ },s.

dmaidaniuk:
Thanks, @Daniel-Dos , it seems that this script updated long time ago. Mentioned Jersey version in it is 2.21.1. Nowadays we have version 2.26.x. However we can use this script as template for updating procedure.

Regards,
Dmytro

Daniel-Dos:
Hi @dmaidaniuk , 

I agree, this is very old. Another note is that this version 2.21.1 is also in the main [pom.xml](https://github.com/mvc-spec/ozark/blob/master/pom.xml) of the repository.

chkal:
Hey everyone,

I updated the Jersey dependency to the latest version:

https://github.com/mvc-spec/ozark/pull/114

All tests run fine. So I merged this one.

Then I updated the integration test setup to execute tests against the latest Glassfish nightly build which does include Jersey 2.26-b07.

https://github.com/mvc-spec/ozark/pull/115

Unfortunately tests started to fail. So I didn't merge it for now. I'll try to find some time to have a deeper look at the test failures later.

chkal:
It looks like jersey/jersey#3582 completely breaks our validation code. See my comment on #115 for details. Actually it may be possible that our code relied on the Jersey bug which was just fixed. Not sure about the details and to be honest: I don't want to spend much more time on debugging this weird `ValidationInterceptor` contract.

So the question is how to proceed. My idea would be to do the following:

* Don't completely remove `BindingInterceptorImpl` but just remove the part related to validation errors. This way we can still try to use the old code to handle binding errors until we find a better way to deal with it. At the same time we can experiment with the new validation handling. I think to do so you basically just have to remove the `catch(ConstraintViolationException)` part in the existing `BindingInterceptorImpl` class.
* Update the same file as I did in #115 to let Travis run the integration tests against the latest Jersey version.

This way we can perhaps get a green bar on Travis. :beers:

dmaidaniuk:
Hi @chkal ,

Yes, proposed by you steps could help here. I will try to do them and see what Travis will say.

Regards,
Dmytro

chkal:
Hey @dmaidaniuk. Any news on this issue?

I'm about to start some restructuring of the code (group id, package names, directory names) and therefore it would be great to get the current status of all pending PRs.

dmaidaniuk:
Hi @chkal ,
I didn't spend too much time to check proposed by you merged solution. On my first quick try it looks like combined changes with old BindingInterceptorImpl alongside to new ValidationInterceptor don't work. You can see to this try here:
https://github.com/dmaidaniuk/ozark/tree/validationinterceptor2
I will look to this PR again in couple next days more closer. It could be that I miss something last time there.

Regards,
Dmytro

chkal:
Thanks for the update. My plan for the next days is to do some basic restructuring of the project and also integrate the TCK tests into the build. I hope these additional tests will help us moving things forward without breaking anything.

chkal:
Hey!

I'm currently working on porting this pull request to the new project structure.

First of all: Thank you very much for your work on this. It seems to basically work fine. I'm adjusting a few aspects but your PR was very helpful. So thanks for that.

The only problem is that currently Glassfish 5.0 b25 throws a weird exception if I put a `@ValidateOnExecution` on the controller:

```
javax.validation.ValidationException: Multiple ValidateOnExecution annotation definitions are hosted in the hierarchy of de.chkal.mvc.test.validation.ValidationController#post method.
	at org.glassfish.jersey.server.validation.internal.ValidateOnExecutionHandler.processAnnotation(ValidateOnExecutionHandler.java:186)
	at org.glassfish.jersey.server.validation.internal.ValidateOnExecutionHandler.processMethod(ValidateOnExecutionHandler.java:140)
	at org.glassfish.jersey.server.validation.internal.ValidateOnExecutionHandler.validateMethod(ValidateOnExecutionHandler.java:123)
...
```

But there is no inheritance hierarchy for this controller as there is no base class. Maybe a Jersey bug. Not sure.

But it is working fine if I set validate on execute to `NONE` via `validation.xml`.

Maybe this was causing the tests to fail back then? How did you disable the JAX-RS BV integration in your tests back then?

chkal:
I just created this Jersey issue: jersey/jersey#3659

dmaidaniuk:
Hi @chkal,

I'm glad that my PR is useful for you. Also I just read your comment in Jersey issue #3659 and it seems that your assumption about CDI proxy and annotation duplication is correct. I agree that bug is very weird. Jersey should correctly work with proxies.
Last time when I was running tests for this PR, I didn't noticed this problem with GlassFish 5. I will double-check it with latest nightly build of it
Meanwhile I will also cherry-pick my commits on the top of current master's changes too.

BTW, do you have any idea how to deal with binding errors? If we will solve this problem too than we can fully skip from vendor-specific code for `BindingResultImpl` class.

Regards,
Dmytro

chkal:
Hey @dmaidaniuk,

thanks again for your help with this. No need for you to cherry pick the commits. I guess I'll commit the slightly modified version of your PR in the next 1-2 days. Just want to cleanup some things and make sure all the tests pass.

A few thoughts about topics related to this PR:

* I was thinking about suggesting adding `@InterceptorBinding` to the `@Controller` annotation. I thought that this would eliminate the need for adding a "virtual" binding annotation to attach the interceptor in Ozark. But then I realized that we still need to use this virtual binding annotation because this way we can make sure that the interceptor is only applied to controller methods and not to other types having this annotation (like the JAX-RS providers)
* I think we should add the binding annotation not to the controller class but to individual methods. This allows us control the validation on a more fine-grained level. This is especially important for mixed JAX-RS / Controller classes where only a single method is a controller and other methods are regular JAX-RS resource methods.
* Currently the `ValidationInterceptor` doesn't validate parameters binding on controller fields. That's something that is required by the spec and was done by the old implementation. But IMO we can address this in a subsequent PR.

I'll ping you as soon as I pushed my branch.

dmaidaniuk:
Hi @chkal,

I see. OK, I will wait for your changes. Also I will think about the last issues from your post with parameters binding on controller fields.

Regards,
Dmytro

chkal:
Looks like there are tests in test suite which verify validation for controller fields. These tests are failing. Therefore I just implemented validation of controller fields which was straight forward. But the tests are still failing. Looks like disabling validation via `@ValidateOnExecution` doesn't work for resource fields in Jersey. 

IMO this is a bug. I just filed another Jersey issue: https://github.com/jersey/jersey/issues/3660

dmaidaniuk:
Hi @chkal ,

It is weird, but looks like the reference implementation of JAX-RS has completely wrong integration with Bean Validation spec. :)

chkal:
Yeah, it is really weird. I'm still working on getting the tests green. I hope to be able to push something today.

chkal:
I finally merged my branch to master. See my announcement here:

https://groups.google.com/forum/#!topic/ozark-dev/iA73x2PEbYM

Thanks a lot for your hard work on this. I'll close this PR now. Let's continue the discussion on the mailing list.

PR #110, opened 2017-06-11
Replaced VariantSelector with new code that uses JAX-RS's Request class

chkal:
Now I think I found a really good solution for getting rid of `VariantSelector`. Let me repeat why it is currently needed:

If a controller method returns null and uses `@View`, JAX-RS won't resolve the content type correctly. That is because JAX-RS thinks that the user wants to send "no content" if the resource returns null.

My idea to fix this is to basically do the same thing as `VariantSelector`, which is to
1. Read `@Produces` annotations on the controller to get an idea of what the controller can produce.
2. Use a complex algorithm to match these content type(s) against the `Accept-*` headers sent by the client.

Both steps were previously implemented in `VariantSelector`. The nice thing is that the (2) is also provided by JAX-RS's `Request#selectVariant()` method. Therefore, we only need to implement (1). I'm not sure why the original authors didn't do it this way. It is much simpler.

My first attempt so simply remove `VariantSelector` and always use `text/html` as part of #107 broke some tests which seem to test exactly this case. But now all the tests are still passing. Therefore, I think I was able to preserve the original behavior. Hooray! :-)


dmaidaniuk:
Great news! Vendor-independent RI became by one step closer.

chkal:
Merged, because all tests passed and nobody raised any concerns!

PR #111, opened 2017-06-17
Replaced Jersey-specific feature-discovery logic by common JAX-RS implementation

dmaidaniuk:
Ozark is implemented using a number of JAX-RS providers (like a MessageBodyWriter, ContainerRequestFilter, ContainerResponseFilter, etc). These providers are registered with the JAX-RS runtime using a Jersey specific SPIs called AutoDiscoverable/ForcedAutoDiscoverable. However, this custom registration process not supported by other JAX-RS implementations, e.g RESTEasy. Thus we need to re-implement Ozark features discovery in a vendor-independent way.

### DIFF
--- a/ozark/src/main/java/org/glassfish/ozark/jaxrs/OzarkFeature.java
+++ b/ozark/src/main/java/org/glassfish/ozark/jaxrs/OzarkFeature.java
@@ -13,57 +13,63 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.glassfish.ozark.jersey;
+package org.glassfish.ozark.jaxrs;
 
-import org.glassfish.jersey.internal.spi.AutoDiscoverable;
-import org.glassfish.jersey.internal.spi.ForcedAutoDiscoverable;
-import org.glassfish.ozark.binding.BindingInterceptorImpl;
-import org.glassfish.ozark.core.ViewRequestFilter;
-import org.glassfish.ozark.core.ViewResponseFilter;
-import org.glassfish.ozark.core.ViewableWriter;
-import org.glassfish.ozark.locale.LocaleRequestFilter;
-import org.glassfish.ozark.security.CsrfProtectFilter;
-import org.glassfish.ozark.security.CsrfValidateInterceptor;
-
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Priority;
-import javax.mvc.annotation.Controller;
+import javax.mvc.engine.Priorities;
 import javax.servlet.ServletContext;
 import javax.ws.rs.ConstrainedTo;
 import javax.ws.rs.RuntimeType;
 import javax.ws.rs.core.Configuration;
 import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Feature;
 import javax.ws.rs.core.FeatureContext;
-import java.util.Arrays;
-
-import static org.glassfish.ozark.util.AnnotationUtils.getAnnotation;
+import javax.ws.rs.ext.Provider;
+import org.glassfish.ozark.binding.BindingInterceptorImpl;
+import org.glassfish.ozark.core.ViewRequestFilter;
+import org.glassfish.ozark.core.ViewResponseFilter;
+import org.glassfish.ozark.core.ViewableWriter;
+import org.glassfish.ozark.jersey.OzarkModelProcessor;
+import org.glassfish.ozark.locale.LocaleRequestFilter;
+import org.glassfish.ozark.security.CsrfProtectFilter;
+import org.glassfish.ozark.security.CsrfValidateInterceptor;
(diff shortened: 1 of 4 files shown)


chkal:
Hey @dmaidaniuk,

thanks a lot for then new pull request. I've a few comments:

* Could you explain why you implemented a new way to checker whether Ozark needs to be enabled? Doesn't `config.getClasses()` work any more?
* I like that you added more log statements. We really need to log more do simplify debugging.
* Some integration tests started to fail. Could you check what is causing this?

Thanks

dmaidaniuk:
HI @chkal ,

Yes, I saw that some integration tests failed and will take a look what happens there.

Coming back to your question about why not use `config.getClasses()` as I wrote in [one of Google group discussions](https://groups.google.com/d/msg/ozark-dev/pYw5u8VacnA/F1N297vYCwAJ) unfortunately RESTEasy returns in `config` object only classes from sub-packages `org.jboss.resteasy`. It means that implementation provide only own classes there but not a features from application or other 3rd-party libraries as Ozark. I don't know if it is an issue of RESTEasy or such moment not clearly specified by JAX-RS spec.

chkal:
Hey @dmaidaniuk,

> unfortunately RESTEasy returns in config object only classes from sub-packages org.jboss.resteasy. It means that implementation provide only own classes there but not a features from application or other 3rd-party libraries as Ozark.

My first thought was that this seems to be some kind of RESTEasy bug. Then I checked the JAX-RS spec but (as so often) it is quite unclear on what this method should return. If Jersey returns resources classes and RESTEasy not, this is IMO something that should be clarified in the spec.

However, if implementations behave differently here, it is a good idea to not use some other way to discover controllers.

chkal:
After thinking about this a bit more. The javadocs of `FeatureContext#getConfiguration()` states:

> Get a live view of an internal configuration state of this configurable instance. Any changes made using methods of this Configurable instance will be reflected in the returned Configuration instance.

This means that we can some random set of classes here, depending on the order of feature class registrations. So if our feature gets added first, we won't see components added later.

That makes using `configuration.getClasses()` pretty useless for scanning. Especially as Jersey adds resources into this set and RESTEasy only providers and features.

dmaidaniuk:
Hi @chkal , 
Yesterday I took a look for broken test `org.glassfish.ozark.test.annotations.AnnotationsIT` and it is exactly that case which you described in discussion in Google group about disadvantage of `Feature` interface:

> If the Application subclass returns non empty sets for getClasses() and getSingletons(), then @Provider will be ignored and the user has to manually register the Feature implementation in the Application subclass.

Tested REST application has not empty set `getClasses()` and that's why `OzarkFeature` not registered in context. Currently I'm thinking how to solve this issue generally, without some Jersey-specific code.

chkal:
Thanks for the update! Yes, I also ran into this issue. Up till now I've not idea how to solve this. Any ideas are welcome.

chkal:
Hey @dmaidaniuk. Any news on this one. Did you find any portable way to fix this issue?

dmaidaniuk:
Hi @chkal,

Only one approach, which comes into my mind, is try to dynamically create a copy of user's `Application` subclass with extended by our Ozark feature list for method `getClasses()`. I played a bit with library [Byte Buddy](http://bytebuddy.net) and it seems I can create such modified copy of particular class on runtime. So my plan is next:

1. Create mentioned wrapper-class (clone) of user's Application subclass.
2. Disable detection by JAX-RS implementation for original `Application` subclass.
3. Force JAX-RS implementation to load my wrapper-class and as a result enable Ozark feature.

First 2 point was easy. I created a clone of user's `Application` class and extended its method `getClasses()` by Ozark feature. Also I removed annotation `ApplicationPath` from original user's class. This automatically disable its detection by Jersey.
However I can't find proper way how to deal with point №3. Dynamically generated and loaded class is accessible in classloader but somehow it is ignored by JAX-RS and even by Servlets initializer. I will try to find solution for this in next few days. Meanwhile you can take a look to my experiments here: https://github.com/dmaidaniuk/ozark/tree/featurediscovery_exp on branch `featurediscovery_exp`. Maybe you will notice something which I miss or give me some fresh idea there.

Regards,
Dmytro

chkal:
Hey @dmaidaniuk,

thanks a lot for sharing this summary. I see the problem. And I'm not sure if there is a portable way to implement No.3. Especially as application servers typically perform annotation scanning BEFORE any application lifecycle listeners are called. That's because even `ServletContainerInitializer` requires annotation metadata and AFAIK `ServletContainerInitializer` is the first thing that fires.

chkal:
I'll close this PR for now. I doesn't merge any more and there are test failures. We now have JAX-RS implementation specific workaround for Jersey and RESTEasy. See my post on the mailing list regarding the new [RESTEasy module](https://groups.google.com/forum/#!topic/ozark-dev/vA20HbqH_MY).

Of cause a portable solution would be great and any pull request for it is more than welcome. I'm thankful for every single line of code we can remove from `ozark-jersey` and `ozark-resteasy`.

PR #112, opened 2017-06-25
updated jade4j to 1.2.5 version

vsmid:
Minor commit to allow usage of some newer jade4j features. 
This will also remove the need to manually exclude old version of jade4j from ozark-jade extension dependency and adding a new one explicitly for users who want to use newer jade4j features.


chkal:
Sorry for the delay! Thanks a lot! Looks great!

vsmid:
np, thanks😊

PR #115, opened 2017-07-08
Use Glassfish nightly builds for integration tests

chkal:
As Jersey starts to address some of the issues we reported, it makes sense to execute the integration tests against the Glassfish Nightly Builds. This way we get the new Jersey versions as early as possible.

For example: https://github.com/jersey/jersey/issues/3582

The new Jersey version which fixes this issue is available in the nightly but not yet in the promoted build.


chkal:
Looks like we cannot update to the latest Jersey version for now. The validation integration breaks in this case. :unamused: 

For me it looks like the `ValidationInterceptor` contract changed in some weird way. In my tests `ctx.proceed()` doesn't throw BV exceptions any more if `@ValidateOnExection` is disabled, which is in some regard understandable. But this breaks our code and validation completely stops working.

So I'll close this for now.

PR #117, opened 2017-07-11
Update Jar's Extensions.

Daniel-Dos:
I did some upgrades from jar's to extensions.

Note for Freemarker.
  - 2.3.26-incubating - http://freemarker.org/freemarkerdownload.html
- The new Configuration was deprecated = http://freemarker.org/docs/pgui_config_incompatible_improvements.html

{  },s.

### DIFF
--- a/ext/asciidoc/pom.xml
+++ b/ext/asciidoc/pom.xml
@@ -31,7 +31,7 @@
         <dependency>
             <groupId>org.asciidoctor</groupId>
             <artifactId>asciidoctorj</artifactId>
-            <version>1.5.2</version>
+            <version>1.5.4</version>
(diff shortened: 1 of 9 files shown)


chkal:
Thanks a lot! Looks great! I agree that we should try to always use the latest versions of the template engines. All tests are green so I'll merge this now.

Daniel-Dos:
Yes, I agree too. So it was not possible to update the StringTemplate and velocity, it seems that they stopped supporting them.

chkal:
Yeah, Velocity seems to be dead. The latest entry in the "recent news" section of the site has been published in 2010. ;-)

PR #118, opened 2017-07-13
Back to older Glassfish 5 nightly

chkal:
Back to an older Glassfish 5 build which doesn't break `BindingInterceptorImpl`...


chkal:
This GF build seems to work fine... So merging this now...

PR #122, opened 2017-07-29
Update Maven module artifactIds

chkal:
This PR is a first step to restructure the Maven modules of Ozark. It contains the following two commits:

* The module `ozark` is renamed to `ozark-core`. This allows us to add additional modules `ozark-jersey` and `ozark-resteasy` later on which will contain the JAX-RS implementation specific code.
* Currently all the parent POM artifactIds are `parent` which is a bit confusing as you usually only see the artifactId of a module in your IDE. I renamed the artifactIds to `ozark-parent`, `ozark-parent-ext` and `ozark-parent-test` to fix this.

I hope everyone agrees with these changes. Let me know if you have any objections.

### DIFF
--- a/ext/asciidoc/pom.xml
+++ b/ext/asciidoc/pom.xml
@@ -21,7 +21,7 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.mvc-spec.ozark.ext</groupId>
-        <artifactId>parent</artifactId>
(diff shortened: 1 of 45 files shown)


chkal:
Ok, great! Thanks for reviewing!

I've already prepared a draft of the new ozark-jersey module. It needs some cleanup but basically it seems to work fine. Expect another PR soon. :smile:

PR #124, opened 2017-08-17
Created new Jersey module

chkal:
Hey everyone,

this PR introduces the new Ozark Jersey module. The goal is to move all the Jersey dependent code into this module and keep Jersey out of the core module. In the long run we will try to get rid of all JAX-RS implementation specific code, but for now this split will allow us to support other JAX-RS implementations as well. Even if it requires some integration code.

A few note about this PR:
  * I created a new class called `OzarkInitializer` which triggers the framework initialization. There is a `DynamicFeature` in the core module which triggers initialization, but as discussed earlier this won't work if the user specifies providers using `Application.getClasses()`. The Jersey module uses the Jersey specific `ForcedAutoDiscoverable` SPI to trigger the initialization as well. Therefore it will work in all cases for Jersey users, just as it does today.
  * A new SPI called `ConfigProvider` is used to collect all providers required by Ozark. Most of the providers are contributed by the core module via `DefaultConfigProvider`. On top of that the Jersey module adds the Jersey-specific `OzarkModelProcessor` via `JerseyConfigProvider`.
  * This means that 2 of the 3 aspects for which we still need Jersey specific APIs have already been migrated into the Jersey module. The only one left is the binding interceptor.
  * I also had to clean up the OSGi bundle manifest creation a bit which was A LOT OF work as I'm not so familiar with OSGi. However, the setup is now much simpler and deploying Ozark as a Glassfish module works fine. We now use standard `jar` packaging and just use the `maven-bundle-plugin` for generating the manifest file. That works great.

Let me know what you think.


ozark/src/main/java/org/glassfish/ozark/bootstrap/OzarkInitializer.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Arrays;
+import java.util.ServiceLoader;
+import java.util.logging.Logger;
+
+/**
+ * Main class for triggering initialization of Ozark
+ *
+ * @author Christian Kaltepoth
+ */
+public class OzarkInitializer {
```
sputnikci: \[Checkstyle\] ERROR: Utility classes should not have a public or default constructor\.

ozark/src/main/java/org/glassfish/ozark/bootstrap/OzarkInitializer.java
```diff
@@ -0,0 +1,87 @@
...
```
sputnikci: \[Checkstyle\] ERROR: Utility classes should not have a public or default constructor\.

sputnikci: \[Checkstyle\] ERROR: Class OzarkInitializer should be declared as final\.

sputnikci: \[Checkstyle\] ERROR: Class OzarkInitializer should be declared as final\.

sputnikci: \[Checkstyle\] ERROR: Class OzarkInitializer should be declared as final\.

sputnikci: \[Checkstyle\] ERROR: Class OzarkInitializer should be declared as final\.

dmaidaniuk: I agree with @sputnikci  that utility class OzarkInitializer must be defined as final.

chkal: Fixed

Daniel-Dos:
Impressive @chkal , this separation was well organized, I can not wait to see it work with other JAX-RS implementations.

{  },s

chkal:
Merged! Thanks a lot for reviewing..

PR #132, opened 2017-08-31
Implement MvcUriBuilder

lefloh:
Implementation of [MvcUriBuilder][1]. 
Following method of `MvcContext` is still not implemented:

```URI uri(String identifier, List<Object> params);```

Reason is that method overloading is not supported by EL. This is the reason why one of the integration tests is failing. 

I started a [discussion in the mailing list][2] if we should drop this method.

 [1]: https://github.com/mvc-spec/mvc-spec/blob/master/api/src/main/java/javax/mvc/MvcUriBuilder.java
 [2]: https://groups.google.com/forum/#!topic/jsr371-users/VU2BitCkJsQ


chkal:
Merged! Thanks a lot!

Issue #98, opened 2016-05-08
Using Editorconfig


mvcbot:
**Original issue OZARK-82 created by lefloh:**

I’d suggest to use <editorconfig>(http://editorconfig.org/) to ensure that every developer uses the same charset, indention settings etc. Here’s a possible configuration file:

```
# Configuration file for EditorConfig: http://editorconfig.org
root = true

<*>
charset = utf-8
end*of*line = lf
indent_style = space
indent_size = 4
insert*final*newline = true
trim*trailing*whitespace = true

[*.{txt,md}]
trim*trailing*whitespace = false

[*.properties]
charset = latin1

[travis.yml]
indent_size = 2
indent_style = space
```

An example is pushed to [this branch](https://github.com/mvc-spec/ozark/compare/master...lefloh:editorconfig).

PR #133, opened 2017-09-03
Remove overloaded method MvcContext#uri(identifier, params)

lefloh:
Method overloading is not supported by EL. See [this discussion](https://groups.google.com/forum/#!topic/jsr371-users/VU2BitCkJsQ) and [mvc-spec PR #98](https://github.com/mvc-spec/mvc-spec/pull/98).


chkal:
I removed the method manually. I guess I should have merged this one before to prevent conflicts. :smile:

Issue #135, opened 2017-09-06
Check why Ozark doesn't work as a Glassfish module any more


chkal:
Ozark should actually work if installed as a module in Glassfish. Previously we ran our test suite against Glassfish 5.0 b09 and it worked fine. After updating to Glassfish 5.0 b25 it doesn't work any more. Therefore I disabled the corresponding build profile:

https://github.com/mvc-spec/ozark/commit/20abe384ea34253727614c024b7ea50d46214005

It looks like the initialization doesn't work any more. Any help on this is welcome. Let me know if you need more details.

jeyvison:
I can take a look at this one

chkal:
That's awesome! Let me know if you need more details.

jeyvison:
@chkal , actually can you specify the steps so i can try to reproduce it here?

chkal:
I think the best way to understand how to reproduce the problem is to have a look at the travis build script. See:

https://github.com/mvc-spec/ozark/blob/39498ca4d223ca26625e66644c8af1e9ce92b3d9/.travis/tests.sh#L21-L31

Basically Ozark is built and then installed as a module in Glassfish. Then the test apps are deployed without Ozark bundled in `WEB-INF/lib`.

For some reason this fails with more recent Glassfish builds. I didn't test 5.0.Final, but the latest promoted builds failed in Travis.

jeyvison:
Just tested with the final version  of glassfish and no error is thrown, but maven complains about the inexistence of the **Integration** profile. Is this profile supposed to exist?

jeyvison:
Forget about the integration profile, that was my bad ˆˆ. I didn't include all the test wars in the deploy folder but now it`s running all the tests successfully. Maybe we should change the ${GLASSFISH_URL} to download the newest version of glassfish.

chkal:
In this case you could try to send a pull request to remove the `#` from this to reactivate the integration tests... Let's see if Travis builds are fine in this case:

https://github.com/mvc-spec/ozark/blob/39498ca4d223ca26625e66644c8af1e9ce92b3d9/.travis.yml#L7-L10

You could also update the Glassfish download URL to the 5.0 final release:

https://github.com/mvc-spec/ozark/blob/39498ca4d223ca26625e66644c8af1e9ce92b3d9/.travis/tests.sh#L5

chkal:
[...]

PR #141, opened 2017-09-29
reactivate integration tests on glassfish

jeyvison:
Reactivate integration tests as talked in #135


chkal:
Wow! Awesome! Looks like it works fine with the final builds. That's great! I wonder why it works now. May be related to the changes we committed recently. Anyway! Thank you so much for taking care of this.

PR #142, opened 2017-10-29
Updated Velocity core library version to 2.0

dmaidaniuk:


Daniel-Dos:
Quite Cool @dmaidaniuk , you managed to make it work without relying on Velocity Tools.

dmaidaniuk:
Hi @Daniel-Dos ,

Yeah, I just followed by suggestion from @chkal and moved class `WebappResourceLoader` directly into extension project. :)

Daniel-Dos:
Very Good @dmaidaniuk  .

Thanks. : )

chkal:
This looks great! And all the tests passed! Great work.

However, I think we should switch to the Velocity Tools dependency after the new version has been released so that we get future updates of this class.

I'll merge this now. Thanks.

dmaidaniuk:
Hi @chkal ,

Definitely we need to switch back to `velocity-tools-view` when its version 3.0 will be released.

Regards,
Dmytro

PR #144, opened 2017-11-20
Moved Viewable class from the spec to Ozark

chkal:
This PR contains the required changes for Ozark regarding mvc-spec/mvc-spec#109.

Please note that the tests will fail until the spec PR is merged.


chkal:
Merging this now, as the corresponding spec issue is now resolved.

PR #145, opened 2017-11-22
Add Jetbrick View engine.

Daniel-Dos:
I found this template engine by browsing the sparkjava page.
This template is very similar to Velocity, so I found it interesting to have it in Ozark.

http://subchen.github.io/jetbrick-template/2x/overview.html

https://github.com/subchen/jetbrick-template-2x

Thanks.

### DIFF
--- a/ext/jetbrick/src/main/java/org/mvcspec/ozark/ext/jetbrick/JetbrickViewEngine.java
+++ b/ext/jetbrick/src/main/java/org/mvcspec/ozark/ext/jetbrick/JetbrickViewEngine.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright © 2017 Ivar Grimstad (ivar.grimstad@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mvcspec.ozark.ext.jetbrick;
+
+import java.io.IOException;
+import java.io.Writer;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.mvc.engine.ViewEngineContext;
+import javax.mvc.engine.ViewEngineException;
+import javax.servlet.ServletContext;
+
+import org.mvcspec.ozark.engine.ViewEngineBase;
+
+import jetbrick.template.JetEngine;
+import jetbrick.template.JetTemplate;
+import jetbrick.template.TemplateException;
+import jetbrick.template.web.JetWebEngine;
+
+/**
+ * @author Daniel Dias
+ */
+@ApplicationScoped
+public class JetbrickViewEngine  extends ViewEngineBase {
+
+	private  JetEngine jetEngine;
+
+	@PostConstruct
+	public void init() {
+		jetEngine = JetWebEngine.create(servletContext);
+	}
+
+	@Inject
+	private ServletContext servletContext;
+
+	@Override
+	public boolean supports(String view) {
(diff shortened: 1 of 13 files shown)


chkal:
Hey @Daniel-Dos. That's awesome! I'll have a closer look at it in the next days! Thanks a lot! :beers:

Daniel-Dos:
you are welcome @chkal . :smiley:

ext/jetbrick/src/main/java/org/mvcspec/ozark/ext/jetbrick/JetbrickViewEngine.java
```diff
@@ -0,0 +1,60 @@
...
+	private ServletContext servletContext;
+
+	@Override
+	public boolean supports(String view) {
+		return view.endsWith(".jetx");
+	}
+
+	@Override
+	public void processView(ViewEngineContext context) throws ViewEngineException {
+		jetEngine = JetWebEngine.create(servletContext);
```
chkal: What do you think about creating the engine once. Looks like this **could** be an expensive operation and therefore creating it once could make sense. We could simply use a `@PostConstruct` method for this and convert the engine into a field.

test/jetbrick/src/main/webapp/WEB-INF/glassfish-web.xml
```diff
@@ -0,0 +1,28 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<!DOCTYPE glassfish-web-app PUBLIC "-//GlassFish.org//DTD GlassFish Application Server 3.1 Servlet 3.0//EN" "http://glassfish.org/dtds/glassfish-web-app_3_0-1.dtd">
+<glassfish-web-app error-url="">
```
chkal: Do we really need this? I know that some of the tests also include this file, but actually we are starting to run the tests against other containers, so it would be great to prevent container-specific deployment descriptors.

ext/jetbrick/src/main/java/org/mvcspec/ozark/ext/jetbrick/JetbrickViewEngine.java
Daniel-Dos: Hi @chkal ,

in case it should look like this:

``` java
private JetEngine jetEngine;

@PostConstruct
public void setup() {
     jetEngine = JetWebEngine.getEngine();
     jetEngine = JetWebEngine.create(servletContext);
    }
```

test/jetbrick/src/main/webapp/WEB-INF/glassfish-web.xml
Daniel-Dos: I can remove this descriptors without problems.

ext/jetbrick/src/main/java/org/mvcspec/ozark/ext/jetbrick/JetbrickViewEngine.java
chkal: Hmmm. What's the difference between `JetWebEngine.getEngine()` and `JetWebEngine.create(servletContext)`? Looks redundant to me.

Daniel-Dos: I removed this ```JetWebEngine.getEngine ()``` and functioned normally. You will not need this instruction.

from the documentation I saw this section:

>In Web integration mode, JetEngine is created by JetWebEngine.create (...) (singleton mode), so that we can get JetEngine

had also seen as an example in an extension to Struts 2 that uses it in line 43: 

https://github.com/subchen/jetbrick-template-2x/blob/master/jetbrick-template-struts/src/main/java/jetbrick/template/web/struts/JetTemplateResult.java

chkal: Ok, great! So just push the changes and we can merge the PR. Thanks.

Daniel-Dos: done. thank you.

chkal:
Looks good to me. Thanks a lot! Merging this now!

Daniel-Dos:
Thanks.

PR #149, opened 2017-12-30
Fixed NPE in case if ServletContext injection not works

dmaidaniuk:
Fixed NPE in case if ServletContext injection not works for particular JAX-RS implementation


chkal:
Looks good. Thanks a lot.

Issue #100, opened 2016-05-09
Upgrade to thymeleaf 3.0.0


mvcbot:
**Original issue OZARK-84 created by eddumelendez:**

Thymeleaf 3 was announced and there are some changes to make mvc compatible with this.

Pull Request:

https://github.com/mvc-spec/ozark/pull/9

PR #152, opened 2018-02-11
Change redirect status code from 302 (Found) to 303 (See other)

lefloh:
See MVC Spec [#100 Clarify POST-redirect-GET status code](https://github.com/mvc-spec/mvc-spec/issues/100).

### DIFF
--- a/test/redirect/src/main/java/org/mvcspec/ozark/test/redirect/RedirectController.java
+++ b/test/redirect/src/main/java/org/mvcspec/ozark/test/redirect/RedirectController.java
@@ -55,7 +55,7 @@ public class RedirectController {
     @Path("response2")
     public Response getResponse2() {
         eventReceived = false;
-        return Response.status(Response.Status.FOUND)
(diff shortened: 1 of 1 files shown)


chkal:
Not sure if we should change this. I don't fully understand why this test was added, but maybe it just verifies, that users can use a different kind of redirect (302 instead of 303) if they manually build the JAX-RS `Response`? And actually [this controller](https://github.com/mvc-spec/ozark/blob/0299508224ed7f684773c8e091f44f9691169f82/test/redirect/src/main/java/org/mvcspec/ozark/test/redirect/RedirectController.java#L47-L52) method already sends a 303 redirect, although it is using a slightly different way to create it.

lefloh:
Thanks for the comment! I'm also still unsure why this test was added. But after [the discussion here](https://github.com/mvc-spec/mvc-spec/issues/100#issuecomment-364890036) I think it makes sense to leave an example how to send a 302.
Closing this one.

PR #156, opened 2018-03-08
Force Jersey to skip validation for MVC controllers

chkal:
This pull request will add a custom `ValidationInterceptor` to the Jersey module which will prevent Jersey from performing validation for MVC requests.

Unfortunately this still isn't enough to get rid of the `@ValidateOnExecution(NONE)` requirement because now the Hibernate Validation CDI interceptor will perform the validation. So by default Jersey actually performs manual validation of BV constraints before the Hibernate Validator CDI interceptor gets invoked, so the latter one will never fail because Jersey will abort processing before.

The good news is that if find a solution for the Hibernate Validation interceptor, we won't need  `@ValidateOnExecution(NONE)` for both Jersey and RESTEasy. And that would be AWESOME!

Travis CI, test please! ;-)

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -118,7 +118,7 @@
         <dependency>
             <groupId>javax</groupId>
             <artifactId>javaee-web-api</artifactId>
-            <version>7.0</version>
+            <version>8.0</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


chkal:
According to my tests you won't need `@ValidateOnExecution(NONE)` any more with:

* Glassfish 5
* Tomcat + Jersey
* Tomcat + RESTEasy (if `resteasy-validator-provider-11` is NOT in your dependencies)

Unfortunately you still need `@ValidateOnExecution(NONE)` with:

* Wildfly 12
* Tomcat + RESTEasy (with `resteasy-validator-provider-11` module)

I'll need some more time to fix this too.

Daniel-Dos:
A question, about the validation part in Jersey.

Since you will no longer have to use @ValidateOnExecution (NONE), then you do not need to add this to Jersey as well:

https://github.com/mvc-spec/ozark/blob/2998dd5e239ae4372eca9fd5287c9c9637ff178a/test/validation/src/main/java/org/mvcspec/ozark/test/validation/MyApplication.java#L38

chkal:
Yeah, I think we can get rid of this one! So if you have some time, just create a pull request.

And actually we should be able to get rid of all the `@ValidateOnExceution(NONE)` annotations in the tests.

PR #157, opened 2018-03-13
Skip RESTEasy validation for MVC controllers

chkal:
This pull request will disable the RESTEasy validation for MVC controllers. There are still some TODOs, but it is basically working fine. This means that you won't need `@ValidateOnExecution(NONE)` any more on Wildfly and Tomcat with RESTEasy.

### DIFF
--- a/resteasy/pom.xml
+++ b/resteasy/pom.xml
@@ -31,7 +31,7 @@
     <name>Ozark RESTEasy</name>
 
     <properties>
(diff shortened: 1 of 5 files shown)


chkal:
I forgot to mention: There is one downside of the current implementation. It will require the `resteasy-validator-provider-11` module on the classpath. This is already the case for Wildfly, but if you are running Tomcat with RESTEasy, you will have to manually include it.

PR #159, opened 2018-04-06
Enforce Ozark bootstrapping only in server mode

chkal:
Fix for an issue reported on the user list. Travis, test please.


chkal:
Just saw that Travis already tested this fix in my fork of Ozark. So merging this now.

PR #160, opened 2018-04-12
Allow @MvcBinding on methods

cdollar393:
The first commit adds code to grab annotations from the getter method of a property and return them in addition to annotations found on the field. That gets the `@MvcBinding` annotation working so that it is picked up in ValidationInterceptor, but at this stage if you deploy an application like my example and force a validation error you'll see messages like these in the logs:
`WARNING [org.mvcspec.ozark.binding.validate.ValidationInterceptor] Cannot resolve paramName for violation: ConstraintViolationImpl{interpolatedMessage='size must be between 1 and      16', propertyPath=formPost.arg0.firstName, rootBeanClass=class eu.agilejava.mvc.HelloController$Proxy$_$$_WeldSubclass, messageTemplate='{javax.validation.constraints.Size.message}'}`

That message is coming from the fact that in my example I have the `@FormParam` annotated on the setter method for the property, and the commit above only checked the getter method for annotations. So we need to also check the setter for annotations, and the second commit does that.

As you can see from the code, I'm just doing some simple reflection to guess the getter and setter names. On the setter side, the code is currently expecting that only one `setXXX` method will be found in the bean, otherwise an exception will be thrown. There may be a way to determine the violated property's class from the ConstraintViolation to avoid that bit of guessing for the setter, but I didn't dig in deep enough to know for sure.

If you have any questions just let me know. Or if there are any changes that you'd like for me to make I'm happy to do that as well. I'm sure there are other ways to implement this sort of functionality. Thanks!

### DIFF
--- a/core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
+++ b/core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
@@ -22,7 +22,11 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
```diff
@@ -83,17 +88,20 @@ else if (lastNode.getKind() == ElementKind.PARAMETER && nodes.size() == 2) {
...
 
-            Class<?> leafBeanClass = violation.getLeafBean().getClass();
             Field field = leafBeanClass.getDeclaredField(node.getName());
-
-            return field.getAnnotations();
+            allAnnotations.addAll(Arrays.asList(field.getAnnotations()));
 
         } catch (NoSuchFieldException e) {
-            throw new IllegalStateException(e);
+            // ignore for now
```
chkal: Maybe we should still throw an exception in this case? But I'm not 100% sure. If we cannot find the field we most likely also won't be able to find getters and setters, because in this case that code that extracted the correct `Node` is broken.

core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
```diff
@@ -118,4 +126,85 @@ else if (lastNode.getKind() == ElementKind.PARAMETER && nodes.size() == 2) {
 
     }
 
+    private static Optional<Annotation[]> setterAnnotationsForField(Class<?> leafBeanClass, String fieldName) {
+        String setterName = setMethodName(fieldName);
+        List<Method> possibleSetters = Arrays.stream(leafBeanClass.getDeclaredMethods())
+            .filter(m -> m.getName().equals(setterName))
+            .collect(Collectors.toList());
+
+        if(possibleSetters.isEmpty())
```
chkal: Could you please always use `{}` blocks even if the `if` only contains one statement?

chkal:
Awesome! Thanks a lot for your help with this. No need to hurry. Looking forward to see if using the `java.beans` code works. :-)

cdollar393:
I've created a new PR that uses `java.beans` instead of low-level reflection, and it is here [1]. If you'd prefer that I push back to this branch to use this PR instead just let me know and I can do that. Thanks!

[1] https://github.com/mvc-spec/ozark/pull/162

PR #162, opened 2018-04-17
@MvcBinding on bean methods using java bean API

cdollar393:
Here's attempt number two at getting `@MvcBinding` working on bean methods.

The first commit is nothing more than unit tests for `ConstraintViolations`, based on the superclass and subclass example that I originally posted to the mailing list. If you run the tests at this point they will fail with a `NoSuchFieldException` being thrown from `ConstraintViolations`.

The second commit is the new fix, using the `java.beans` API rather than low-level reflection to discover getter and setter methods and annotations on those methods. The unit tests are all successful after the fix.

Lastly, I wanted to address your comment about potentially throwing an exception when we encounter a `NoSuchFieldException`. I think we do not want to throw an exception in that case, because the field lookup will always fail with a `NoSuchFieldException` in the case of a subclass that only overrides methods from the super class. I created a unit test that illustrates this and put it in this gist [1]. I can add it in to the code base if you want, but since it really does nothing more than verify behavior of the JRE I didn't know if it was worth including.

Let me know if you have any questions or would like for me to make any other changes. I'm happy to help!

[1] https://gist.github.com/cdollar393/7e277029ac06ee5b2af9016d61fb0090

### DIFF
--- a/core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
+++ b/core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
@@ -18,11 +18,19 @@ package org.mvcspec.ozark.binding.validate;
 import javax.validation.ConstraintViolation;
 import javax.validation.ElementKind;
 import javax.validation.Path;
+import java.beans.BeanInfo;
+import java.beans.IntrospectionException;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
+import java.util.Set;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mvcspec/ozark/binding/validate/ConstraintViolations.java
```diff
@@ -118,4 +129,46 @@ else if (lastNode.getKind() == ElementKind.PARAMETER && nodes.size() == 2) {
...
+
+                Method setter = descriptorOpt.get().getWriteMethod();
+                if(setter != null) {
+                    annotationsSet.addAll(Arrays.asList(setter.getAnnotations()));
+                }
+
+            }
+        }
+        catch(IntrospectionException e) {
+            // should we throw an exception here?
```
chkal: Maybe we could just log a warning in this case?

test/constraintViolations/src/main/java/org/mvcspec/ozark/test/constraintViolations/TestAnnotation.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * A test annotation that is taking the place of @MvcBinding for these tests.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.PARAMETER, ElementType.FIELD, ElementType.METHOD })
+@Documented
+public @interface TestAnnotation {
```
chkal: I already merged mvc-spec/mvc-spec#144, so you should be able to use `@MvcBinding` in your test.

test/constraintViolations/src/test/java/com/mvcspec/ozark/test/constraintViolations/ConstraintViolationsTest.java
```diff
@@ -0,0 +1,198 @@
...
+ * <p>
+ * These tests are based on a scenario where we have a number of existing java
+ * bean classes that have bean validation constraints already applied to both
+ * fields and getter methods. We extend those beans so that we can annotate the
+ * field getter methods with {@link TestAnnotation} (which simulates the {@link
+ * javax.mvc.binding.MvcBinding} annotation), and annotate the setter methods
+ * with {@link javax.ws.rs.FormParam}, as would be done on a JAX-RS/MVC {@link
+ * javax.ws.rs.BeanParam}.
+ */
+public class ConstraintViolationsTest {
```
chkal: The test looks great! Thanks a lot for adding this. Having more test coverage is awesome!

But actually the `tests` directory is just for integration tests which contain real web application which can be deployed to server to run asserts from "outside". I know that `tests` directory name isn't optimal, but it was a very early decision to call it this way and we didn't rename it yet.

So feel free to move the tests to `core/src/test/java`..

chkal:
BTW: Please ignore the Travis test failures for now. Looks like they are caused by some other issue.

cdollar393:
No problem! I'm happy to help. I'll make those changes and will push up some new commits later today.

cdollar393:
I just pushed some more commits that should take care of those last couple of issues. If you notice anything else just let me know!

PR #165, opened 2018-04-22
Modifying returned sets and lists to be immutable as defined in MVC Spec

kmohanarangam:
This is a minimal change to return immutable Set/List to satisfy MVC Spec requirement


chkal:
Looks great! Thanks a lot!

PR #172, opened 2018-05-22
remove glassfish-web.xml

Daniel-Dos:
this PR remove the files glassfish-web.xml in the tests the projects.


chkal:
Great! Thanks! I guess we don't need these files at all.

PR #176, opened 2018-06-19
View engine groovy

Daniel-Dos:
- add a new View Egine - Groovy 

Reference - > http://docs.groovy-lang.org/next/html/documentation/template-engines.html

### DIFF
--- a/ext/groovy/src/main/java/org/mvcspec/ozark/ext/groovy/GroovyViewEngine.java
+++ b/ext/groovy/src/main/java/org/mvcspec/ozark/ext/groovy/GroovyViewEngine.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright © 2017 Ivar Grimstad (ivar.grimstad@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mvcspec.ozark.ext.groovy;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.nio.charset.Charset;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.mvc.engine.ViewEngineContext;
+import javax.mvc.engine.ViewEngineException;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+
+import org.codehaus.groovy.control.CompilationFailedException;
+import org.mvcspec.ozark.engine.ViewEngineBase;
+import org.mvcspec.ozark.engine.ViewEngineConfig;
+
+import groovy.lang.Writable;
+import groovy.text.Template;
+import groovy.text.markup.MarkupTemplateEngine;
+
+/**
+ * @author Daniel Dias
+ */
+@ApplicationScoped
+public class GroovyViewEngine extends ViewEngineBase {
+
+    @Inject
+    @ViewEngineConfig
+    private MarkupTemplateEngine markupTemplateEngine;
+
+    @Inject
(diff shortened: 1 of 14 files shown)


ext/groovy/src/main/java/org/mvcspec/ozark/ext/groovy/GroovyViewEngine.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Inject
+	@ViewEngineConfig
+	private MarkupTemplateEngine markupTemplateEngine;
+
+	@Inject
+	private ServletContext servletContext;
+
+	public boolean supports(String view) {
+		return view.endsWith(".tpl");
```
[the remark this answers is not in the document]
Daniel-Dos: Hi, 

I not see in documentation other extension difference of .tpl .

I finding this simple tutorial in site the Spring boot what use this template with that extension :

https://spring.io/blog/2014/05/28/using-the-innovative-groovy-template-engine-in-spring-boot

Daniel-Dos:
Hi @chkal, 

please checking the review .

thanks .

test/groovy/src/test/java/org/mvcspec/ozark/test/groovy/GroovyIT.java
```diff
@@ -0,0 +1,56 @@
...
+    @After
+    public void tearDown() {
+        webClient.closeAllWindows();
+    }
+
+    @Test
+    public void testView1() throws Exception {
+        final HtmlPage page = webClient.getPage(webUrl + "resources/hello?user=mvc");
+        final Iterator<HtmlElement> it = page.getDocumentElement().getHtmlElementsByTagName("h1").iterator();
+        assertTrue(it.next().asText().contains("mvc"));
```
chkal: Usually `assertTrue()` assertions aren't a good choice, because error messages aren't very expressive in case of failures. They will only tell you something like "You expected true but it was false". And this won't point you to the root cause of the failure.

Instead try using something like:

```java
assertEquals("Hello mvc", it.next.asText().trim());
```

Or:

```java
assertThat(it.next.asText(), cOREMatchers.containsString("Hello mvc"));
```

This will give you errors like: "Expected 'Hello mvc' but got 'Hello $name'", which is much better to identify the root cause.

ext/groovy/src/main/java/org/mvcspec/ozark/ext/groovy/GroovyViewEngine.java
```diff
@@ -0,0 +1,78 @@
...
+        Map<String, Object> model = new HashMap<>(context.getModels());
+        model.put("request", context.getRequest(HttpServletRequest.class));
+        Charset charset = resolveCharsetAndSetContentType(context);
+        try (Writer writer = new OutputStreamWriter(context.getOutputStream(), charset);
+            InputStream resourceAsStream = servletContext.getResourceAsStream(resolveView(context));
+            InputStreamReader in = new InputStreamReader(resourceAsStream, "UTF-8");) {
+            Template template = markupTemplateEngine.createTemplate(in);
+            Writable output = template.make(model);
+            output.writeTo(writer);
+        } catch (IOException | CompilationFailedException | ClassNotFoundException e) {
```
chkal: I'm just wondering: Which method is throwing `ClassNotFoundException` here? Or can we remove this exception type here?

Daniel-Dos: I see this . 

this line :  markupTemplateEngine.createTemplate(in); and what throws the exception .

In this case, I can create a separate catch for this exception.

chkal: No, that's fine. The exception looks weird, but it the `MarkupTemplateEngine` throws that exception it is fine to catch it together with the other ones.

Daniel-Dos: Do you think that would be good?   

` } catch (IOException | CompilationFailedException e) {
            throw new ViewEngineException(e);
        }  catch (ClassNotFoundException e) {
			e.printStackTrace();
		}`

test/groovy/src/test/java/org/mvcspec/ozark/test/groovy/GroovyIT.java
Daniel-Dos: Most of the tests are with this assertTrue, do you recommend switching to assertEquals or assertThat?

ext/groovy/src/main/java/org/mvcspec/ozark/ext/groovy/GroovyViewEngine.java
ivargrimstad: The original approach where you catch them all and re-throw as VewEngineException is better. This one would just swallow it and you wouldn't know what happened...

Daniel-Dos: I'll do the update for The original approach .

PR #178, opened 2018-06-20
resolve cdi-named-beans in thymeleaf

### LINKED ISSUE #64: provide a way to leave @Named bean available on other template engines
**Original issue OZARK-46 created by turini:**

It'd be nice to automatically leave CDI named beans accessible on other template engines like freemarker, mustache and so on -- when using ozark extensions, just like ozark-freemark.

gtudan:
This adds support for `@Named`-Annotated beans in Thymeleaf.
At first I tried extending the WebContext, but that doesn't work as most of it's methods are final.

The Context enumerates all named beans upon constructions, which should be a fairly cheap operation. Initialization is delayed until the reference is actually used in the template. The original idea for this came from https://github.com/inbuss/thymeleaf-cdi, so @inbuss should get the credit for it. 

About testing: I haven't seen any test in the ext package, so I'm not sure how to approach this. Do you have any plans for this?

Closes #64

### DIFF
--- a/ext/thymeleaf/src/main/java/org/mvcspec/ozark/ext/thymeleaf/CDIWebContext.java
+++ b/ext/thymeleaf/src/main/java/org/mvcspec/ozark/ext/thymeleaf/CDIWebContext.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright © 2017 Ivar Grimstad (ivar.grimstad@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mvcspec.ozark.ext.thymeleaf;
+
+import org.thymeleaf.context.IWebContext;
+import org.thymeleaf.context.LazyContextVariable;
+
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.Any;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.util.AnnotationLiteral;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Thymeleaf context for resolving variables from the model or cdi context. In addition to the functions provided by
+ * {@link org.thymeleaf.context.WebContext} this allows to resolve beans annotated with {@linkplain javax.inject.Named}
+ * from the cdi context.
+ * <p>
+ * Beans will be reolved in this order:
+ * <ol>
+ *     <li>special variables (like "request")</li>
+ *     <li>variables from the model</li>
+ *     <li>named beans from the cdi context</li>
+ * </ol>
+ * </p>
+ *
+ * @author Gregor Tudan
+ */
+class CDIWebContext implements IWebContext {
+
+    private final HttpServletRequest request;
+    private final HttpServletResponse response;
+    private final ServletContext context;
+    private final Locale locale;
+
+    private final BeanManager beanManager;
+    private final Set<String> cdiNamedBeans;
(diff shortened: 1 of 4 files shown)


ext/thymeleaf/src/main/java/org/mvcspec/ozark/ext/thymeleaf/CDIWebContext.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public HttpServletResponse getResponse() {
+        return response;
+    }
+
+    @Override
+    public HttpSession getSession() {
+        return request.getSession();
```
chkal: I think we can use `request.getSession(false)` here... It looks like `null` is allowed in this case... 

See the [API docs of IWebContext](https://www.thymeleaf.org/apidocs/thymeleaf/3.0.0.RELEASE/org/thymeleaf/context/IWebContext.html#getSession--).

ext/thymeleaf/src/main/java/org/mvcspec/ozark/ext/thymeleaf/CDIWebContext.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    private Object getCdiBean(String name) {
+        return new LazyContextVariable<Object>() {
+            @Override
+            protected Object loadValue() {
+                Bean<?> bean = beanManager.getBeans(name).iterator().next();
+                CreationalContext ctx = beanManager.createCreationalContext(bean);
+                return beanManager.getReference(bean, bean.getBeanClass(), ctx);
```
chkal: To be honest, I'm not sure if we create a memory leak here. I ran into a similar issue with one of our client projects recently where we used `Instance<SomeBean>` for lazily resolving CDI beans which weren't release correctly.

I guess this can only happen for `@Dependent` scoped beans, but as this is the default if you forget about the annotation, it may be dangerous. Maybe we can find a way keep track of the `CreationalContext` instances and then call `CreationalContext#release()` to destroy the dependent-scoped instances?

Or maybe I'm wrong and there won't be a leak? Not completely sure. It is still too early in the morning. ;-)

gtudan: Quiet possible - I need to do some reading on this. I remember seeing memory leaks on non-contextual objects (when passing null as creational context), but in this case I would expect CDI to handle the beans as defined by their context.

The thymeleaf-context does not have any lifecycle hooks, so we might need to find another way of knowing when the request has finished.

ext/thymeleaf/src/main/java/org/mvcspec/ozark/ext/thymeleaf/CDIWebContext.java
gtudan: Right, done

ext/thymeleaf/src/main/java/org/mvcspec/ozark/ext/thymeleaf/CDIWebContext.java
chkal: > but in this case I would expect CDI to handle the beans as defined by their context.

Sure, but if there isn't a scope annotation it will be `@Dependent` scoped and therefore depend on the lifecycle of the class of the injection point. But in our case there isn't an injection point as we are doing the lookup manually. But as I said, I'm also not 100% sure about this.

> The thymeleaf-context does not have any lifecycle hooks, so we might need to find another way of knowing when the request has finished.

We could simply define a `close()` method on our context class and call it after the template has been evaluated in a `finally` block. I guess this should work.

gtudan: Ah, I get it now  - you are right, the dependent scope probably won't work and most things I read point out that we'll have to take care of these beans by disposing the creational context.

I'll take a shot at it tomorrow.

chkal: Awesome! Thanks a lot! :beers:

gtudan: It was easier than I thought, seems to work fine.
I also noticed that thymeleaf does not flush buffers after rendering, so I did that before dropping into the disposing to improve the initial site response. The user will get the site rendered but the loading indicator will keep going a little longer in case a bean takes longer to dispose.

gtudan:
Damn, guess I was too quick on this - I'll check what's wrong with the integration tests.

gtudan:
Okay, I got it sorted - I got the import in the test wrong and endet up with the JSF-RequestScoped instead of the CDI-one. Funny enough wildfly didn't seem to care, but any other app-server broke.

Should be fine now.

chkal:
Ok, I'll merge this now! Thanks @gtudan for helping with this.

Issue #177, opened 2018-06-20
Log a WARNING in case a MVC controller is annotated with @Named 


rsoika:
As we have already discussed [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jsr371-users/-yH5KnsK88U/Vpvl3Gy0AwAJ) it does not make sense to annotate a MVC Controller class with `@Named` and a scope. 

The following conroller code works fine in Wildfly/RestEasy:

```
@Named
@RequestScoped
@Controller
@Path("/connection")
public class LoginController implements Serializable {
  String id;
  public String getId() {
	 return id;
  }
  public void setId(String id) {
	this.id = id;
  }
....
}
```

The property 'id' can be used in the view 

```
...
  <h1>#{loginController.id}</h1>
...
```

But the same code will not work for Glassfish/Jersey!

The reason is, that Jersey (or the CDI container) news the controller when rendering the view and so the properties are lost. ReastEasy (or the CDI Contianer) re-uses the bean also during the rendering phase (which makes sense as it is annotated as RequestScoped). 
But as I understand this behavior is not precisely defined by the jaxrs spec. So a implementation may re-new the controller or not. 

Of course both implementations are working fine in case a RequestScoped CDI bean is injected into the MVC Controller. I think this is exactly the same as when injecting the mvc Models Bean.

As Christian already noted in the usergroup, the JAX-RS 2.1 spec states the following in 3.1.1 "Lifecycle and Environment":

_By default a new resource class instance is created for each request to that resource. First the constructor (see Section 3.1.2) is called, then any requested dependencies are injected (see Section 3.2), then the appropriate
method (see Section 3.3) is invoked and finally the object is made available for garbage collection. An implementation MAY offer other resource class lifecycles, mechanisms for specifying these are outside the scope of this specification. E.g. an implementation based on an inversion-of-control framework may support all of the lifecycle options provided by that framework._

So for me it is now clear that the mvc controller annotated with `@Named` is not working in all Implementations (even if it works in Wildfly). My suggestion would be - if this is possible - to log a warning when an MVC controller is annotated with `@Named` or any kind of Scope.

This will help developers to not fall into this trap. In my case, the problem first occurred when I tried to deploy my application in Glassfish. And I have to refactor the complete application.

chkal:
Just to quickly share my thoughts on this.

[...]

Issue #161, opened 2018-04-17
Check if controllers are CDI beans


chkal:
The spec states:

> Unlike in JAX-RS [...], MVC classes are REQUIRED to be CDI-managed beans only"

Ozark should check that and fail if a controller is NOT a CDI beans. This could happen for example if the application is an implicit CDI archive in `annotated` discovery mode and the controller hasn't been defined with `@RequestScoped` or other scope annotations.

mscharhag:
Do you have an idea on how to get the CDI discovery mode at runtime? I don't think CDI provides an easy way to get this information.

chkal:
Not sure if this is possible at all using the standard CDI API.

However, I don't think that this is necessary to fix this issue. Basically we could just check whether the controller is a CDI bean in a `ContainerRequestFilter` which has access to the resource class. We could even use `ViewRequestFilter` for this. In this case we could simply ask the `BeanManager` for the corresponding class, which should always give us a result if the controller is a CDI bean. If the `BeanManager` doesn't know about the resource class, it won't be a CDI bean.

Does this make sense?

mscharhag:
Yes, you are right. Checking the BeanManager should be enough.

I am trying to look into this. However, I'm a bit confused.

I created a sample application with `annotated` discovery mode and a controller class without any scope annotation. If I run this application with glassfish/jersey an exception is thrown before `ViewRequestFilter` is processed. So checking it there would be too late.

```
org.glassfish.hk2.api.UnsatisfiedDependencyException: There was no object available for injection at SystemInjecteeImpl(requiredType=Models,parent=TestController,qualifiers={},position=-1,optional=false,self=false,unqualified=null,1007007569)
	at org.jvnet.hk2.internal.ThreeThirtyResolver.resolve(ThreeThirtyResolver.java:75)
	at org.jvnet.hk2.internal.Utilities.justInject(Utilities.java:1012)
	at org.jvnet.hk2.internal.ServiceLocatorImpl.inject(ServiceLocatorImpl.java:1008)
	at org.glassfish.jersey.inject.hk2.AbstractHk2InjectionManager.inject(AbstractHk2InjectionManager.java:231)
	at org.glassfish.jersey.inject.hk2.ImmediateHk2InjectionManager.inject(ImmediateHk2InjectionManager.java:54)
	at org.glassfish.jersey.ext.cdi1x.internal.AbstractCdiBeanSupplier$2.getInstance(AbstractCdiBeanSupplier.java:112)
[...]

PR #182, opened 2018-07-07
Checks if controllers are CDI beans (#161) and logs a warning when annotated with @Named (#177)

mscharhag:
As discussed in #161.

### DIFF
--- a/core/src/main/java/org/mvcspec/ozark/servlet/OzarkServletContextListener.java
+++ b/core/src/main/java/org/mvcspec/ozark/servlet/OzarkServletContextListener.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright © 2017 Ivar Grimstad (ivar.grimstad@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.mvcspec.ozark.servlet;
+
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+import javax.servlet.annotation.WebListener;
+
+import java.util.Set;
+import java.util.logging.Logger;
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mvcspec/ozark/servlet/OzarkServletContextListener.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Throws an exception if the passed controller class is not present in the CDI BeanManager.
+     * For example, this can happen with glassfish / Jersey if the application is using CDI discovery mode "annotated"
+     * and the controller class is not annotated with any scope annotation.
+     * Unlike in JAX-RS resources, MVC controllers are required to be CDI-managed beans.
+     */
+    private void failIfNoCdiBean(Class<?> controllerClass) {
+        Set<Bean<?>> controllerBeans = beanManager.getBeans(controllerClass);
+        if (controllerBeans == null || controllerBeans.isEmpty()) {
+            String message = String.format("The controller %s is not a managed CDI bean. Maybe the controller class is" +
```
chkal: Minor issue: There is a space missing at the end of the first string literal. So the error message will end up like this:

> Maybe the controller class **ismissing** a scope annotation

core/src/main/java/org/mvcspec/ozark/servlet/OzarkServletContextListener.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    /**
+     * Logs a warning if the passed controller class is annotated with {@link Named}.
+     */
+    private void warnIfAnnotatedWithNamed(Class<?> controllerClass) {
+        Named namedAnnotation = controllerClass.getAnnotation(Named.class);
+        if (namedAnnotation != null) {
+            String message = String.format("Controller class %s is annotated with @Named. Typically this should not be " +
```
chkal: Perhaps adjust the error message a bit. "Unexpected results" is actually not quite correct. Perhaps something like:

> Controller class %s is annotated with `@Named`. Typically this is not be required, because you should never access the controller directly from a view.

chkal:
Further reviews welcome!

core/src/main/java/org/mvcspec/ozark/servlet/OzarkServletContextListener.java
```diff
@@ -0,0 +1,91 @@
...
+     * Throws an exception if the passed controller class is not present in the CDI BeanManager.
+     * For example, this can happen with glassfish / Jersey if the application is using CDI discovery mode "annotated"
+     * and the controller class is not annotated with any scope annotation.
+     * Unlike in JAX-RS resources, MVC controllers are required to be CDI-managed beans.
+     */
+    private void failIfNoCdiBean(Class<?> controllerClass) {
+        Set<Bean<?>> controllerBeans = beanManager.getBeans(controllerClass);
+        if (controllerBeans == null || controllerBeans.isEmpty()) {
+            String message = String.format("The controller %s is not a managed CDI bean. Maybe the controller class is " +
+                    "missing a scope annotation (e.g. @RequestScope).", controllerClass.getName());
```
lefloh: Typo: `@RequestScoped` is missing the 'd'.

chkal: Good catch

Issue #151, opened 2018-02-11
Add support for a default file extension for view names


chkal:
See the following discussion on the mailing list:

https://groups.google.com/forum/#!topic/jsr371-users/3ky4bw_8A0I

I think we should try to implement a minimal version of this requirement in Ozark without mentioning it ins the spec. This makes it an Ozark specific feature which we may  add to the spec later on.

I could imagine a configuration parameter like this:

```java
properties.put(Properties.DEFAULT_VIEW_FILE_EXTENSION, "jsp");
```
And some code similar to this one to apply this default (pseudo code):

```java
if( !viewName.endsWith( "." + defaultViewFileExtension) ) {
    viewName = viewName + "." + defaultViewFileExtension;
}
```

dmaidaniuk:
Hi @chkal,

I'm taking this task for implementation. Will provide pull request soon.

Regards,
Dmytro

chkal:
Thanks a lot, @dmaidaniuk. Looking forward to your pull request.

Daniel-Dos:
Hello,

I have tried to do this issue as mentioned above, I do not know if it is really this, I am committing to my changes and would like some help on this:

1 - I created a new attribute in the Properties class : 

https://github.com/Daniel-Dos/Ozark-integrationTests/blob/master/core/src/main/java/org/mvcspec/ozark/Properties.java#L38

2 - I created a method to return a jsp as default if there was no change of the same via JAx-rs getProperties: 

https://github.com/Daniel-Dos/Ozark-integrationTests/blob/master/core/src/main/java/org/mvcspec/ozark/OzarkConfig.java#L64

3 - I did this to check the possible return of views:

https://github.com/Daniel-Dos/Ozark-integrationTests/blob/master/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java#L154-L162

chkal:
Thanks @Daniel-Dos. I'll have a deeper look at the code as soon as possible and provide some feedback.

dmaidaniuk:
Thank you @Daniel-Dos! 
Could you make pull request with your all changes so it will be suitable to review them in one place, please? Did you also extend tests for this new functionality?

Regards,
Dmytro

chkal:
@Daniel-Dos You could also simply create a pull request and add "[WIP]" in front of the description. This will simplify reviewing and adding comments to the code. What do you think?

Daniel-Dos:
Hello,

[...]

PR #204, opened 2018-09-17
Add support for a default file extension for view names

dmaidaniuk:
See an issue https://github.com/mvc-spec/ozark/issues/151 for details.

### DIFF
--- a/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
+++ b/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
@@ -191,7 +191,7 @@ public class ViewResponseFilter implements ContainerResponseFilter {
         }
 
         String resultView = viewName;
-        if (!viewName.contains(".") && !viewName.contains(":")) {
+        if (!viewName.contains(".")) {
             resultView += "." + defaultExtension;
         }
         return resultView;
--- a/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
+++ b/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
@@ -183,7 +183,8 @@ public class ViewResponseFilter implements ContainerResponseFilter {
      * Append to view name default extension if one available and applicable.
      */
     protected static String appendExtensionIfRequired(String viewName, String defaultExtension) {
-        if (viewName == null || viewName.startsWith(REDIRECT) || defaultExtension == null) {
+        if (viewName == null || viewName.startsWith(REDIRECT)
+            || defaultExtension == null || defaultExtension.length() == 0) {
             return viewName;
         }
 
--- a/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
+++ b/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
@@ -15,6 +15,7 @@
  */
 package org.mvcspec.ozark.core;
 
+import org.mvcspec.ozark.OzarkConfig;
 import org.mvcspec.ozark.event.AfterControllerEventImpl;
 import org.mvcspec.ozark.event.ControllerRedirectEventImpl;
 
@@ -80,6 +81,7 @@ import static org.mvcspec.ozark.util.PathUtils.*;
 public class ViewResponseFilter implements ContainerResponseFilter {
 
     private static final String REDIRECT = "redirect:";
+    private static final String DOT_CHAR = ".";
 
     @Context
     private UriInfo uriInfo;
@@ -96,6 +98,9 @@ public class ViewResponseFilter implements ContainerResponseFilter {
     @Inject
     private Messages messages;
 
+    @Inject
+    private OzarkConfig ozarkConfig;
+
     @Override
     public void filter(ContainerRequestContext requestContext,
                        ContainerResponseContext responseContext) throws IOException {
@@ -111,6 +116,7 @@ public class ViewResponseFilter implements ContainerResponseFilter {
 
         final Method method = resourceInfo.getResourceMethod();
         final Class<?> returnType = method.getReturnType();
+        final String defaultExtension = ozarkConfig.getDefaultViewFileExtension();
 
         // Wrap entity type into Viewable, possibly looking at @View
         Object entity = responseContext.getEntity();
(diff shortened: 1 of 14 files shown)


dmaidaniuk:
@chkal,
Travis build for this PR failed due to reached timeout for `glassfish-module` job which took more than 49 min to execute. All this time was spent to download artifacts from Maven repository. Do you know why this job executed so slowly? Other jobs in this build executed much faster.

chkal:
@dmaidaniuk You can ignore these timeouts for now. That's a Travis issue. I'm already in contact with the Travis team and they are working on a fix.

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
```diff
@@ -80,6 +81,7 @@
 public class ViewResponseFilter implements ContainerResponseFilter {
 
     private static final String REDIRECT = "redirect:";
+    private static final String DOT_CHAR = ".";
```
chkal: Not sure how others feel about that, but defining a constant for the dot character isn't really required. I would be fine with inlining it.

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
```diff
@@ -111,6 +116,7 @@ public void filter(ContainerRequestContext requestContext,
 
         final Method method = resourceInfo.getResourceMethod();
         final Class<?> returnType = method.getReturnType();
+        final String defaultExtension = ozarkConfig.getDefaultViewFileExtension();
```
chkal: Perhaps we should let `appendExtensionIfRequired` lookup the default extension instead of doing this here? This way we won't need to pass this into every call to `appendExtensionIfRequired`. I think that's fine because effectively `appendExtensionIfRequired` will only be called one per request so there won't be any performance befit of reading the value here.

The only argument for doing the lookup of the default value outside of `appendExtensionIfRequired` is that it simplifies unit testing. But we could work around this by doing something like this:

```java
// call this method in the filter
private static String appendExtensionIfRequired(String viewName) {
  return appendExtensionIfRequired(viewName, ozarkConfig.getDefaultViewFileExtension());
}

protected static String appendExtensionIfRequired(String viewName, String defaultExtension) {
  // same as in your pull request
}
```

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
```diff
@@ -173,6 +179,24 @@ public void filter(ContainerRequestContext requestContext,
...
+        if (viewName == null || viewName.startsWith(REDIRECT)
+            || defaultExtension == null || defaultExtension.length() == 0) {
+            return viewName;
+        }
+
+        String resultView = viewName;
+        if (!viewName.contains(DOT_CHAR) && !viewName.contains(":")) {
+            resultView += DOT_CHAR + defaultExtension;
+        } else if (viewName.endsWith(DOT_CHAR)) {
+            throw new IllegalArgumentException("Invalid format for view name: " + viewName);
```
chkal: I don't think that we should fail here. Of course view names with trailing dots are weird, but technically they will work. So why not removing this else block?

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
```diff
@@ -173,6 +179,24 @@ public void filter(ContainerRequestContext requestContext,
...
+     * Append to view name default extension if one available and applicable.
+     */
+    protected static String appendExtensionIfRequired(String viewName, String defaultExtension) {
+        if (viewName == null || viewName.startsWith(REDIRECT)
+            || defaultExtension == null || defaultExtension.length() == 0) {
+            return viewName;
+        }
+
+        String resultView = viewName;
+        if (!viewName.contains(DOT_CHAR) && !viewName.contains(":")) {
```
chkal: Could you explain the `!viewName.contains(":")` here? Not sure if this is really needed. Especially because I know at least one view engine which supports view names like `react:some-folder/foobar.jsp`.

chkal:
Oh, I forgot to mention one thing. I think I was able to fix the Travis errors you were seeing in #207. But there is no real need to rebase to master just to get this fix. I can restart the failed Travis builds manually if required and usually this fixes the problem with the build. But rebasing world of course also work. :-)

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
dmaidaniuk: Exactly for view engines like mentioned I skipped the default extension attaching in case if the character `:` available in the incoming view name. So, view names like `react:some-folder/foobar.jsp` or `react:some-folder/foobar` will be excluded from handling and remain unchanged.
Do you think that it is an incorrect assumption and we still need to process such view names?

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
dmaidaniuk: Fixed.

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
dmaidaniuk: Done.

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
dmaidaniuk: Else block removed.

dmaidaniuk:
Hi @chkal,

I fixed almost all issues, which you mentioned in review. Only one left here: https://github.com/mvc-spec/ozark/pull/204#discussion_r219689624
I added my explenation for it. Please take a look.
Also, build on Travis passed now after your fix.

Regards,
Dmytro

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
chkal: I don't see a reason why we should not apply the default file extension mechanism if the view name contains `:`. Actually a view name like `react:foobar.jsp` would mean something like "use the react view engine but render the view using a default JSP view engine". So why not apply the default mechanism in this case?

Of course `redirect:` needs special treatment, but this is already done in the first `if` block. So I think it is safe to add the default extension if there is no `.` char in the view name. So I would prefer to simply remove the `!viewName.contains(":")` part.

chkal:
Awesome! Thank you very much for applying the changes. I added a comment to the last remaining one.

core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
dmaidaniuk: OK, I got it. I made the required changes.

dmaidaniuk:
@chkal I made the last requested change. Build on Travis passed.

chkal:
Awesome! Thanks @dmaidaniuk for all your help with this. I'll merge this now! :beers:

dmaidaniuk:
@chkal you're welcome. Glad to help you with this task. I guess that original issue #151 could be also closed now.

Daniel-Dos:
Very good. 

thanks @dmaidaniuk  and @chkal  . 🍻

PR #210, opened 2018-09-29
Add a test suite module

gtudan:
create a test suite module and migrate the thymeleaf example to it


gtudan:
I took a first shot at the test suite module. It's still a quiet large PR, please let me know if you would like to further slim it down. 

Currently only one container (wildly) - I will add more in further steps.

testsuite/pom.xml
```diff
@@ -0,0 +1,216 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.mvc-spec.ozark</groupId>
+        <artifactId>ozark-parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>testsuite</artifactId>
```
chkal: Could you also add a `<name>` element just like in the other modules? Something like:

```xml
<name>Ozark Testsuite</name>
```

testsuite/pom.xml
```diff
@@ -0,0 +1,216 @@
...
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>javax</groupId>
+            <artifactId>javaee-web-api</artifactId>
+            <version>7.0</version>
```
chkal: I think we should use `8.0` here. But I guess you copied the version from some other place in Ozark, correct? In this case let's leave it as it is and extract some common Maven property for the version later on.

testsuite/src/test/java/org/mvcspec/ozark/test/ThymeleafIT.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @ArquillianResource
+    private URL baseURL;
+
+    @Drone
+    private WebDriver webDriver;
+
+    @Deployment(testable = false, name = "thymeleaf")
+    public static Archive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class)
```
chkal: That looks great! But I wonder if we should create some util class which simplifies building a "common base archive". I did this for the TCK, see here:

https://github.com/mvc-spec/mvc-tck/blob/master/tests/src/main/java/org/mvcspec/tck/tests/mvc/models/BuiltinEngineModelTest.java#L48

The implementation is simple, as it delegates all the work to a custom archive builder:

https://github.com/mvc-spec/mvc-tck/blob/b79c16bab5cb8c035a1415d55c5551d67ddfd776/tests/src/main/java/org/mvcspec/tck/util/Archives.java#L28-L34

And a custom builder with specialized methods to build MVC archives:

https://github.com/mvc-spec/mvc-tck/blob/master/tests/src/main/java/org/mvcspec/tck/util/WebArchiveBuilder.java

It is not really required to have something like this in the first PR, but I guess it would make sense to do this shortly after merging it, so we have an easy job of migrating the other examples.

chkal:
I'll merge this now! Thank you so much for your help. I'm happy to see we are finally getting a Arquillian-based test suite! :beers:

Issue #211, opened 2018-10-06
File Extension not work with Viewable


Daniel-Dos:
when trying to use the file extension with the return type Viewable a 500 is displayed as shown in the screenshots: 

![eeror](https://user-images.githubusercontent.com/8139890/46568238-cd43bf00-c917-11e8-8b82-73c540d66c3a.png)

![error2](https://user-images.githubusercontent.com/8139890/46568241-d16fdc80-c917-11e8-8bc2-130cc4926ea7.png)

I made a small correction in the following PR, where to add another else if to Viewable, with that it started working for the 4 return types.

chkal:
Good catch! And thanks for also creating the pull request to fix this issue. :+1:

PR #212, opened 2018-10-06
fix extension to Viewable.

Daniel-Dos:
see my commnets in #211

### DIFF
--- a/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
+++ b/core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
@@ -32,6 +32,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.Produces;
 import javax.ws.rs.ServerErrorException;
+import javax.ws.rs.client.ResponseProcessingException;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerResponseContext;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mvcspec/ozark/core/ViewResponseFilter.java
```diff
@@ -147,6 +147,12 @@ public void filter(ContainerRequestContext requestContext,
                 throw new ServerErrorException(messages.get("EntityToStringNull", resourceInfo.getResourceMethod()), INTERNAL_SERVER_ERROR);
             }
             responseContext.setEntity(new Viewable(view), null, responseContext.getMediaType());
+        } else if (entityType == Viewable.class) {
```
chkal: This will definitely work. But I think we can make the code a bit simpler. Basically this `if` block is almost the same as the `if` block above. The only difference is, that this block gets the view name from `Viewable.getView()` and the block above by calling `entity.toString()`. The rest of the code is exactly the same. 

So may we can merge these blocks into a single `else` block which gets the view name like this:

```java
final String view = appendExtensionIfRequired(
  entityType == Viewable.class ? ((Viewable) entity).getView() : entity.toString()
);
```

This class is getting more and more complex to reason about, so simplifying it is a good thing.

Do you think this can work? Or am I missing something?

Daniel-Dos: This way you put it was much simpler, but the class is complex.

chkal:
Thanks, this looks great! :beers:

PR #219, opened 2018-11-28
migrate mvc-example to test-suite

gtudan:
Hi Christian,

here's another baby-step for the test suite. I migrated another example app (MVC) which took quiet  lot of refactoring for the test suite. I took your ArchiveBuilder from the TCK and modified it for the test suite, so we can add additional libraries (like thymleaf) for some tests that need them.

Additional dependencies dependencies and their version can be configured in the dependency management block of the POM.

The actual test case is pretty straight forward.


Daniel-Dos:
very good  .

chkal:
Awesome! This looks great! I think using the ArchiveBuilder is a great step forward, because it will simplify writing new tests!

Thanks! You are awesome! :beers:

